PR #25, opened 2017-10-10
Task deploy

vinyll:


vinyll:
Need to merge for sharing with the team

PR #26, opened 2017-10-10
saving profile with flask security

vinyll:
- API access requires user to be logged in
- default `Model` is taken from playhouse to gain access to model signals
- profile can be saved properly with the new version based on Flask Security


vinyll:
Need to merge for sharing the project on the staging with the team

cafebabel/core/models.py
```diff
@@ -43,7 +47,7 @@ class UserRoles(db.Model):
 
 class UserProfile(db.Model):
     name = CharField(null=True)
-    user = ForeignKeyField(User)
+    user = ForeignKeyField(User, related_name='profiles', on_delete='CASCADE')
 
     def __str__(self):
-        return self.name if self.name else str(self.user)
+        return self.name or str(self.user.email)
```
davidbgk, line 53: I think the `str()` is now useless?

PR #28, opened 2017-10-17
Submit article unittest

vinyll:
Passed too much time on it, just made it work with _unittest_ 🚎


vinyll:
We went for the pytest version https://github.com/cafebabel/cafebabel.com/pull/29

PR #29, opened 2017-10-17
Submit article

vinyll:
Debugging with non magic (aka `unittest`) helped understand how to get this solved. 🚎 
Still quite not understanding, but at least it's working.

### DIFF
--- a/settings.py
+++ b/settings.py
@@ -11,7 +11,7 @@ SECURITY_CONFIRMABLE = False
 SECURITY_REGISTERABLE = True
 SECURITY_RECOVERABLE = True
 SECURITY_CHANGEABLE = True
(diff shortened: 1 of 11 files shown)


settings.py
```diff
@@ -24,3 +24,19 @@
...
+MAIL_SERVER = 'localhost'
+MAIL_PORT = 25
+MAIL_USE_TLS = False
+MAIL_USE_SSL = False
+MAIL_DEBUG = DEBUG
+MAIL_USERNAME = None
+MAIL_PASSWORD = None
+MAIL_DEFAULT_SENDER = None
+
+EDITORS_EMAIL_DEFAULT = 'editors@cafebabel.com'
```
davidbgk: I would remove emails from github/dev and add a prod config on the server

davidbgk (approved):

PR #33, opened 2017-10-26
Submit article

vinyll:


settings.py
```diff
@@ -37,7 +37,9 @@
 # Prevent writing clear email for avoiding robots on Github sources.
 EDITORS_EMAIL_DEFAULT = '@'.join(['editors', 'cafebabel.com'])
 EDITOR_EMAILS = {
-    'en': '@'.join(['en', 'cafebabel.co.uk']),
-    'fr': '@'.join(['fr', 'cafebabel.fr']),
-    'it': '@'.join(['it', 'cafebabel.it']),
+    'fr': '@'.join(['redaction', 'cafebabel.com']),
```
davidbgk, line 40: Not related to that PR but rethinking about these emails-related stuff: I think we should configure with a catch-all email for now at least to get feedback on the current website without spamming anyone.

davidbgk (approved):
I don't really get where is the article submission though 😆

vinyll:
indeed, the name of the PR is inadequate.
I agree on having a catch all for website feedbacks. Probably a "contact us" link somewhere

PR #34, opened 2017-10-26
wrapped form fields with a class for easy tuning

vinyll:
This renders HTML such as the following:

```html
<div class="text">
    <label for="email">Email Address</label> <input id="email" name="email" type="text" value="">
  </div>
  
  <div class="password">
    <label for="password">Password</label> <input id="password" name="password" type="password" value="">
  </div>
  
  <div class="checkbox">
    <label for="remember">Remember Me</label> <input id="remember" name="remember" type="checkbox" value="y">
  </div>
``` 
Using `<div>` as wrapper as they are semantically meaningless.


johangiraud (approved):
cool, easier like that

PR #39, opened 2017-10-31
creating sections to structure the page

johangiraud:
dunno how to do that without having these containers. I am open to suggestions in case you have some


templates/profile.html
```diff
@@ -7,98 +7,105 @@
 <h1>Create a profile</h1>
 {% endif %}
 
-<ul>
-  <li>Full name: <span data-edit-name=name data-edit-attrs="required">{{ user.profile.name or '' }}</span></li>
-  <li>Website: <span data-edit-name=website data-edit-attrs="type=url"><a href="{{ user.profile.website }}" target=_blank>{{ user.profile.website }}</a></span></li>
-</ul>
+<main>
```
vinyll: I suppose that main should be in the base layout and eventually the `<h1>` title inside of it.

templates/profile.html
```diff
@@ -7,98 +7,105 @@
...
+        .catch(e => alert(e.error))
       })
-      .catch(e => alert(e.error))
-    })
-  </script>
-{% endif %}
+    </script>
+  {% endif %}
+  </section>
+  <section>
```
vinyll: Should this  be `<aside>`?

templates/profile.html
```diff
@@ -7,98 +7,105 @@
...
+    <h2>Contributions</h2>
+  </section>
+</main>
```
vinyll: We want the `<script>` at the end of the document.

vinyll (changes requested):

templates/profile.html
johangiraud: sure

templates/profile.html
johangiraud: to me `aside` means that is a major column and a minor one : `aside`. Here it does't make sense cause both columns have the same importance. What do you think ?

templates/profile.html
johangiraud: ok, why don't you place this `<script>` in an external file like script.js or utils.js ?

vinyll: This page is unique and the script is specific.
We could therefore place the script in a `profile.js` file if you'd rather have it outside.

templates/base.html
```diff
@@ -38,5 +38,7 @@
 
 {% include "security/_messages.html" %}
 {% block body %}{% endblock %}
-
+    
+<main>
```
vinyll: Taking the opening tag also mean you should take the closing tag along.

vinyll (changes requested):

templates/profile.html
johangiraud: yo no prob if the script is used only in this page, it makes sense then

vinyll (approved):

templates/base.html
johangiraud: 👍

PR #44, opened 2017-11-03
Adding tests for API and login/logout

davidbgk:


cafebabel/tests/test_users.py
```diff
@@ -26,3 +27,17 @@ def test_admin_has_all_roles(admin):
...
+def test_authenticated_user_can_access_login_required_page(client, user):
+    login(client, user.email, 'secret')
+    response = client.get('/profile/')
+    assert response.status_code == HTTPStatus.OK
+    assert b'<h1>testy@example.com\'s profile</h1>' in response.data
+
+
+def test_logout_user_cannot_access_login_required_page(client, user):
+    login(client, user.email, 'secret')
+    logout(client)
```
vinyll, line 41: why login then logout?

conftest.py
```diff
@@ -16,6 +16,7 @@ def pytest_runtest_teardown():
 
 @pytest.fixture
 def app():
+    myapp.config['WTF_CSRF_ENABLED'] = False
```
vinyll: Could you also modify the name of the DB so that we don't lose the manual data when running tests?

vinyll (approved):
r+wc

 🆒 for the tests! 🌮

cafebabel/tests/test_users.py
davidbgk: To make sure that logout is effective

PR #48, opened 2017-11-09
added article information in article detail page

vinyll:


cafebabel/tests/test_articles.py
```diff
@@ -108,12 +108,26 @@ def test_access_article_with_large_slug_should_return_200(client, article):
     assert response.status_code == HTTPStatus.OK
 
 
-def test_published_article_should_display_content(client):
-    article = Article.objects.create(title='My title', body='Read me',
-                                     status='published', language='en')
+def test_published_article_should_display_content(client, user):
+    article = Article.objects.create(title='My title', summary='Sum this',
```
davidbgk: Why not the `article` fixture?

cafebabel/core/helpers.py
```diff
@@ -46,3 +55,9 @@ def decorated_view(*args, **kwargs):
                   'An editor is required to perform this action.')
         return func(*args, **kwargs)
     return decorated_view
+
+
+@app.template_filter()
+def reading_time(text):
```
davidbgk: Remove that one or the other one?

templates/articles/detail.html
```diff
@@ -1,23 +1,33 @@
 {% extends "base.html" %}
 
+{% block meta_title %}{{ article.title }}{% endblock %}
+{% block meta_description %}{{ article.summary }}{% endblock %}
+
 {% block body %}
-<article>
+<article lang={{ article.language }}>
```
davidbgk: Better on the `<html>` element?

templates/articles/detail.html
```diff
@@ -1,23 +1,33 @@
...
   {% endif %}
 
   {{ article.body|markdown }}
+
   <p>
-    Published on <time>{{ article.creation_date.date() }}</time>
+    Published on <time>{{ article.creation_date.date() }}</time> by {{ article.author }}
   </p>
+  <nav class=share>
+    <a href="https://twitter.com/share?url={{ request.url }}&text={{ article.title }}&via=cafebabel">Twitter</a>
```
davidbgk: Shouldn't you urlencode the `request.url`?

cafebabel/core/helpers.py
```diff
@@ -46,3 +55,9 @@ def decorated_view(*args, **kwargs):
...
+    words = do_wordcount(text)
+    return ceil(words / 250)
```
davidbgk: Should be tested.

cafebabel/articles/views.py
```diff
@@ -86,7 +86,7 @@ def draft_create():
         return redirect(url_for('draft.draft_detail', draft_id=article.id))
     else:
         return redirect(url_for('article.article_detail', slug=article.slug,
-                                id=article.id))
+                                article_id=article.id))
```
davidbgk, line 89: Missing test here, let's add one!

templates/articles/detail.html
vinyll: Would it make sense to have one on HTML and one on the article if the language of the page and the language of the article were different? (more a general question as it may not apply here)

templates/articles/detail.html
vinyll: :+1:

templates/articles/detail.html
davidbgk: If it happens that would be weird for the visitor (as in my blog :p), maybe we should try to avoid that situation with forced translation of the interface too?

cafebabel/articles/views.py
vinyll: added

PR #59, opened 2017-11-14
published articles are visible in the author's page

vinyll:


settings.py
```diff
@@ -39,6 +39,7 @@
 SECURITY_SEND_CONFIRMATION_TEMPLATE = 'auth/confirmation.html'
 EXPLAIN_TEMPLATE_LOADING = False
 
+DEBUG_TB_ENABLED = False
```
davidbgk: I see why you did that and I tried (very!) hard to do that only for tests with no success 😢

cafebabel/tests/test_profile.py
```diff
@@ -9,3 +10,20 @@ def test_confirm_user_creates_default_profile(app):
     with app.app_context():
         confirm_user(user)
     assert user.profile.name == user.email
+
+
+def test_author_profile_has_no_list_of_drafts(client, user, article):
+    login(client, user.email, 'secret')
+    response = client.get(f'/profile/{str(user.id)}/')
```
davidbgk: I don't think you need the extra `str()`

cafebabel/tests/test_profile.py
```diff
@@ -9,3 +10,20 @@ def test_confirm_user_creates_default_profile(app):
...
+    assert response.status_code == 200
+    assert article.title not in response.get_data(as_text=True)
+
+
+def test_author_profile_has_list_of_published_articles(client, user, article):
+    login(client, user.email, 'secret')
+    article.status = 'published'
+    article.author = user
+    article.save()
+    response = client.get(f'/profile/{str(user.id)}/')
```
davidbgk: Same here.

davidbgk (changes requested):

settings.py
vinyll: I'll suggest another PR with that

davidbgk (approved):

PR #60, opened 2017-11-14
Deal with translations

davidbgk:
More than the translations feature described on #20 it's a proposition for a couple of other things:

* using inheritance in models
* having dedicated folder per kind (should be done for draft for consistency)
* the coupling urls/views/permissions is a bit different to what has been done for now
* separating templates too
* the responsibility of the model for cleaning internal data
* the responsibility of the model for returning its current URL

As always, I'm open to all comments :)


vinyll:
We can be agnostic of the source language from where the translation is from, therefore I was wondering what made you choose to store the `translated_from`?

davidbgk:
We need a reference to the initial article, I wasn't able to find a better name but I'm open to suggestion, I agree that it's confusing :/

cafebabel/translations/models.py
```diff
@@ -0,0 +1,53 @@
+from flask import url_for
+from mongoengine import signals, ValidationError
+
+from .. import app, db
+from ..users.models import User
+from ..articles.models import Article
+
+
+class Translation(Article):
```
vinyll: `translations.models.Translation` sounds very generic while we are being as specific as the translation of an article.
Would it make sense to give a more specific name?

Same question for the indenpendant package. if is so related to the _articles_ package that I'm wondering why extracting it.

We will have translations soon for the system, maybe we don't want the names to be confusing. (Or should we merge the _translations_ module and make it the website translation module? In which case the `class Translation` would still be too specific) 🤔

davidbgk: 1. on translation specificity, I think we should keep it and name the translation of the interface internationalization (i18n) to make the difference
2. for the independence, I found the `articles` folder already too crowded, when you have more than one blueprint in views it's a sign _to me_, I even want to go deeper and extract drafts too with their own set of blueprint/views/templates

For instance here: https://github.com/cafebabel/cafebabel.com/blob/master/templates/articles/edit.html#L4-L13 To me again that rang a bell, sometimes it's less costly to repeat ourselves a bit vs. complexity of the logic (and probably more to come).

vinyll: I agree on point 2.
In order to keep items splitted and yet consistently connected, what do you think of making _drafts_, _translations_ (and current _articles_?) subpackages of the _articles_ package?
In which case we would import _Translation_ from _articles_ and that would still be readable to me ^^

If you'd rather this with 1st level packages I'd suggest just renaming the _Translation_ model subclass _ArticleTranslation_ for the sake of explicity.

cafebabel/tests/test_articles.py
```diff
@@ -102,9 +110,9 @@ def test_published_article_should_offer_social_sharing(client, article):
 
 
 def test_visitor_cannot_change_editor_nor_author(client, editor, user):
-    draft = Article.objects.create(title='My draft', body='Content',
-                                   language='en', status='draft',
-                                   author=user, editor=editor)
+    draft = Article.objects.create(title='My draft', summary='Summary',
```
vinyll: use _article_ fixture?

cafebabel/tests/test_translations.py
```diff
@@ -0,0 +1,222 @@
...
+
+from flask.helpers import get_flashed_messages
+
+from ..articles.models import Article
+from ..translations.models import Translation
+from .utils import login
+
+
+def test_translation_query_should_retrieve_all(app, article, translation):
+    french = app.config['LANGUAGES'][1][0]
```
vinyll: Wondering about the dependency of the settings order here 🤔

cafebabel/translations/models.py
```diff
@@ -0,0 +1,53 @@
...
+    translator = db.ReferenceField(User, required=True)
+    translated_from = db.ReferenceField('Article', required=True)
```
vinyll: I wouldn't mind brainstorming on the name of this field.
Some thoughts:  _article_, _original_, , _source_…

cafebabel/translations/views.py
```diff
@@ -0,0 +1,64 @@
...
+            abort(HTTPStatus.BAD_REQUEST, str(e))
+        flash('Your translation was successfully created.')
+        return redirect(translation.detail_url)
+
+    return render_template('translations/create.html',
+                           article=article, language=language)
+
+
+@blueprint.route('/<regex("\w{24}"):id>/')
+def translation_detail(id):
```
vinyll: Do we need to repeat the _translation_ prefix as this is isolated and _blueprinted_?

docs/architecture/decisions/0004-deal-with-translations.md
```diff
@@ -0,0 +1,39 @@
...
+Each translation must keep a link to the original article.
+
+The review process remains the same with a draft status.
+
+The translated article must both have a redactor and a translator.
+
+
+## Decision
+
+A new MongoDB document will be created for each translation.
```
vinyll, line 27: This makes me wonder about [EmbeddedDocument vs seperated document](http://openmymind.net/Multiple-Collections-Versus-Embedded-Documents/).

vinyll:
And BTW, this PR is a 💣 
👍

cafebabel/translations/models.py
davidbgk: I'll go for moving the folder… which makes me thinking about URLs :)

What about `article/translation/…`? (and same for draft and proposal)

vinyll: That would definitely make sense to me indeed. 👍

settings.py
```diff
@@ -1,6 +1,7 @@
 from pathlib import Path
 
 DEBUG = True
+TESTING = True
```
davidbgk: TODO: should be useless when merging from master with dedicated testing setting file.

davidbgk: Actually, no. I think it's related to the way we load app's extension in the global `__init__` before we set the testing settings.

cafebabel/articles/models.py
```diff
@@ -22,9 +34,22 @@ class Article(db.Document):
...
+        'queryset_class': ArticleQuerySet
+    }
+
     def __str__(self):
         return self.title
 
+    @property
+    def detail_url(self):
+        if self.is_draft:
+            return url_for('draft.detail', draft_id=self.id)
```
vinyll: Out of this scope and I know it's been a short while now, but I always wonder why we pass `draft_id` or `article_id` and not just `id` just like we do with `slug`.

cafebabel/articles/models.py
```diff
@@ -33,6 +58,16 @@ def is_draft(self):
     def is_published(self):
         return self.status == 'published'
 
+    @property
+    def is_translation(self):
+        return False
+
+    def is_translated_in(self, language):
```
vinyll, line 65: 👍

cafebabel/articles/models.py
```diff
@@ -33,6 +58,16 @@ def is_draft(self):
...
+        from .translations.models import Translation  # TODO: circular :/
+        return bool(Translation.objects
+                    .filter(original_article=self, language=language)
+                    .only('id').first())
```
vinyll: `.filter(…).count()` instead?

cafebabel/articles/proposals/views.py
```diff
@@ -0,0 +1,33 @@
+from flask import Blueprint, flash, redirect, render_template, request, url_for
+from flask_mail import Message
+
+from ... import app, mail
+
+blueprint = Blueprint('proposal', __name__)
+
+BODY_EMAIL_TEMPLATE = '''
```
vinyll: Good point!

cafebabel/articles/translations/models.py
```diff
@@ -0,0 +1,52 @@
...
+        super().clean()
+        try:
+            translations = Translation.objects.filter(
+                original_article=self.original_article, language=self.language)
+            if (translations and
+                    (len(translations) > 1 or translations[0].id != self.id)):
+                raise ValidationError('Existing translation already exists.')
+        except Translation.DoesNotExist:
+            pass
+        try:
```
vinyll: Do you think we could replace all this with [`unique_with` from MongoEngine](http://docs.mongoengine.org/apireference.html#mongoengine.base.fields.BaseField)?

cafebabel/articles/translations/views.py
```diff
@@ -0,0 +1,66 @@
...
+blueprint = Blueprint('translation', __name__)
+
+
+@blueprint.route('/new/', methods=['get', 'post'])
+@login_required
+def create():
+    original_article = request.args.get('original')
+    language = request.args.get('lang')
+    if not original_article:
+        abort(HTTPStatus.NOT_FOUND, 'You must specify a valid article.')
```
vinyll: `BAD_REQUEST` instead?

cafebabel/articles/translations/views.py
```diff
@@ -0,0 +1,66 @@
...
+    if not language or language not in dict(app.config['LANGUAGES']):
+        abort(HTTPStatus.NOT_FOUND, 'You must specify a valid language.')
```
vinyll: same question

vinyll (commented):
I only did a part of the review but it might be already interesting to get some feedback

cafebabel/tests/test_translations.py
davidbgk: It concerns me too, especially because that's now hardcoded in a few more places. Maybe it shouldn't be (only?!) a setting?

docs/architecture/decisions/0004-deal-with-translations.md
davidbgk: Honestly, I don't know Mongo enough to choose wisely…

settings.py
davidbgk: Done for mail, cannot be done for debug toolbar though without a [better organization of the project](https://github.com/stephane/flask-skeleton/blob/master/skeleton/__init__.py).

cafebabel/articles/translations/views.py
```diff
@@ -0,0 +1,66 @@
...
+    article = Article.objects.get_or_404(id=original_article)
+
+    if request.method == 'POST':
+        data = request.form.to_dict()
+        try:
+            translation = Translation.objects.create(
+                translator=current_user.id,
+                original_article=article.id,
+                status='draft',
+                editor=article.editor,
```
vinyll, line 25: After a talk with @cafebabel-webmaster, we don't need to store the editor at all.
And the editor from the original article would rarely be the same editor from the translation anyways.

cafebabel/articles/translations/views.py
```diff
@@ -0,0 +1,66 @@
...
+        return redirect(translation.detail_url)
+
+    return render_template('articles/translations/create.html',
+                           article=article, language=language)
+
+
+@blueprint.route('/<regex("\w{24}"):id>/')
+def detail(id):
+    translation = Translation.objects.get_or_404(id=id, status='draft')
+    return render_template('articles/translations/detail.html',
```
vinyll: detail.html displays a draft of translation?
It's not called for displaying a translation but article/detail.html is.

davidbgk: Yes, it is for drafts

cafebabel/articles/translations/views.py
```diff
@@ -0,0 +1,69 @@
...
+def create():
+    original_article = request.args.get('original')
+    language = request.args.get('lang')
+    if not original_article:
+        abort(HTTPStatus.BAD_REQUEST, 'You must specify a valid article.')
+    if not language or language not in dict(app.config['LANGUAGES']):
+        abort(HTTPStatus.BAD_REQUEST, 'You must specify a valid language.')
+    article = Article.objects.get_or_404(id=original_article)
+
+    if request.method == 'POST':
```
vinyll: We should try to be uniform on this.
Eventually we remove the other views where _get_ and _post_ methods are different routes and make a check in the view if you'd rather go this road on others files too.

templates/articles/detail.html
```diff
@@ -31,7 +33,49 @@ <h1>{{ article.title }}</h1>
   </nav>
   {% endblock %}
   {% if current_user.has_role('editor') and article.is_published %}
-    <a class=button href={{ url_for('article.article_edit_form', article_id=article.id) }}>Edit this article</a>
+    <p>
+      <a class=button href={{ url_for('articles.edit', article_id=article.id) }}>Edit this article</a>
+    </p>
+  {% endif %}
+
+  {% if article.original_article %}
```
vinyll: `article.is_translation`

cafebabel/articles/translations/views.py
davidbgk: Isn't it what I did?

cafebabel/tests/test_translations.py
vinyll: The test is just testing a second language. Eventually we could rename it different than `french`?

cafebabel/articles/translations/views.py
vinyll: You did it this way here, but we have it the other way in the other views.

cafebabel/articles/views.py
```diff
@@ -1,180 +1,80 @@
...
-    except (Article.DoesNotExist, mongoengine.errors.ValidationError):
-        abort(HTTPStatus.NOT_FOUND, 'No article found with this id.')
+def delete(article_id):
+    article = Article.objects.get_or_404(id=article_id)
     article.delete()
     flash('Article was deleted.', 'success')
     return redirect(url_for('home'))
+
+
+@articles.route('/to_translate/')
```
vinyll: underscore in the url?

cafebabel/tests/test_articles.py
```diff
@@ -324,3 +227,80 @@ def test_delete_inexistent_article_should_return_404(client, editor, article):
...
+def test_access_published_article_should_link_translations(client, article,
+                                                           translation):
+    article.status = 'published'
+    article.save()
+    translation.status = 'published'
+    translation.save()
+    response = client.get(f'/article/{article.slug}-{article.id}/')
+    text = response.get_data(as_text=True)
+    assert response.status_code == HTTPStatus.OK
+    assert '<p>Translate this article in:</p>' in text
```
vinyll, line 241: This is going to break very soon. Just saying so we have in mind when Johan will redo the contents.

cafebabel/articles/translations/views.py
davidbgk: Well, I don't get it, I made it uniform everywhere in that PR.

cafebabel/tests/test_translations.py
```diff
@@ -0,0 +1,237 @@
...
+    assert translation.title == 'Modified title'
+    assert (get_flashed_messages() ==
+            ['Your translation was successfully updated.'])
+
+
+def test_translation_update_empty_summary(client, user, translation):
+    login(client, user.email, 'secret')
+    data = {
+        'title': translation.title,
+        'summary': '',
```
vinyll: Wasn't this mandatory in the end?

conftest.py
```diff
@@ -45,8 +46,21 @@ def admin():
 
 @pytest.fixture
 def article():
+    en = myapp.config['LANGUAGES'][0][0]
```
vinyll: same question as with `french`

vinyll: and I guess if you're using the language just once you could set it direction at line 53. Unless you might have a good reason.

conftest.py
```diff
@@ -45,8 +46,21 @@ def admin():
...
         title='article title',
         summary='summary text',
-        language=myapp.config['LANGUAGES'][0][0],
+        language=en,
         body='body text')
+
+
+@pytest.fixture
+def translation(user, article):
+    fr = myapp.config['LANGUAGES'][1][0]
```
vinyll: I guess at the end of the review I'm ok with it anyways. :)

readme.md
```diff
@@ -40,6 +40,12 @@ re-initializing the database.
 If existing, the file `settings.local.py` at the root of the project will
 override the default `settings.py` configuration.
 
+Do not forget to create the folder dedicated to statics, by default:
```
vinyll, line 43: This should be generated with `make install`.

templates/articles/to_translate.html
```diff
@@ -0,0 +1,40 @@
...
+      {% endif %}
+    {% endfor %}
+  </ul>
+
+  <ul>
+  {% for article in articles %}
+    <li>
+      <article>
+        <h2>{{ article.title }}</h2>
+        <ul>
```
vinyll: I suppose a `<nav>` would be more appropriate but we see with @cafebabel-webmaster when implementing the frontend.

templates/articles/translations/create.html
```diff
@@ -0,0 +1,21 @@
+{% extends "base.html" %}
+
+{% block body %}
+<h1>Creating a translation in {{ language }} for article “{{ article.title }}”</h1>
```
vinyll, line 4: Just being curious here, are these English quotes?

templates/articles/translations/create.html
```diff
@@ -0,0 +1,21 @@
...
+
+<form method=post action=>
```
vinyll: Is `action=` with no value valid, without even a dot?

vinyll (commented):
Been as quick as I could :/

cafebabel/articles/views.py
davidbgk: No strong opinion, would you remove it?

cafebabel/tests/test_articles.py
davidbgk: I know, and I can't find a better way. I prefer to update the tests when it will be changed. Which reminds me that the CI is not passing on each PR…

cafebabel/tests/test_translations.py
davidbgk: Indeed, actually the test pass because the value when empty is not updated. I remove it.

conftest.py
davidbgk: Updated since then.

readme.md
davidbgk: Do you recommend to run `make install` for devs too?

templates/articles/translations/create.html
davidbgk: Yes :)

templates/articles/translations/create.html
davidbgk: My browser looks to appreciate it. This way it keeps the current query string!

cafebabel/articles/translations/views.py
vinyll: Right, spots I had in mind may have been rewritten. My bad.

cafebabel/articles/views.py
vinyll: I'd use a dash like we do unless you had a reason not to

readme.md
vinyll: Mhh… not sure. Let's keep it like you did for now.

templates/articles/translations/create.html
vinyll: After checking it seems to break the [w3c validator](https://validator.w3.org). although `action=.` should do the job :)
This breaks:

```html
<!doctype html>
<title>a</title>
<form action=>
</form>
```

This passes
```html
<!doctype html>
<title>a</title>
<form action=.>
</form>
```

And [I had this DoD in mind](https://github.com/cafebabel/cafebabel.com/wiki/DoD#pour-le-site) (ok ok, for that point ^^ )

davidbgk: Done 😘

vinyll (approved):

PR #61, opened 2017-11-15
tests have their specific settings file

vinyll:


settings.py
```diff
@@ -39,6 +39,7 @@
 SECURITY_SEND_CONFIRMATION_TEMPLATE = 'auth/confirmation.html'
 EXPLAIN_TEMPLATE_LOADING = False
 
+DEBUG_TB_ENABLED = True
```
davidbgk: I don't think it's useful given that it's enabled by default and shouldn't when debug is set to `False`?

davidbgk (approved):
👍

PR #63, opened 2017-11-16
pitch document form styles + change pitch doc form

johangiraud:


static/css/styles.css
```diff
@@ -400,12 +420,18 @@ select { /* https://stackoverflow.com/questions/3532649/problem-with-select-and-
     appearance: none;
 }
 
-label[for=field-language] {
+label[for=field-language],
+label[for=media-type],
+label[for=format],
+label[for=section] {
```
vinyll: Is there no `class=select` on the wrapper that you do each manually?

templates/articles/proposal.html
```diff
@@ -12,16 +15,41 @@
...
-    <label for=field_name>name *</label><input name=name id=field_name required>
-    <label for=field_email>contact *</label><input type=email name=email id=field_email required>
-    <label for=field_city>city</label><input name=city id=field_city>
+    <label for=field_name>Name *</label><input name=name id=field_name required>
+    <label for=field_email>Contact *</label><input type=email name=email id=field_email required>
+    <label for=field_city>City</label><input name=city id=field_city>
+    <label for=field_topic>Topic in a sentence *<p>ex: Digital nomads in Lisbon</p></label><input name=topic id=field_topic required>
+    <label for=field_angle>Pitch your story<p>ex: I want to meet young people/ participants of the three major digital nomad communities/ events in Lisbon and use their personal stories as an insight to describe the digital nomads landscape in Portugal</p></label><input name=angle id=field_angle>
+    <hr>
+    <label for=media-type>Media
```
vinyll, line 26: Changing these forms fields will require to adapt the backend (views and tests).
David or myself should pair with you to complete this PR.

static/css/styles.css
johangiraud: `select` is inside `label` :-(

templates/articles/proposal.html
johangiraud: ok, this might not be the final-final version, so maybe we do this later

vinyll: ok, but we won't be able to approve/merge it then as it may break the tests.

static/css/styles.css
vinyll: Right. I meant you know how we wrapped `<div class=checkbox>`?
Feel free to do the same here with `<div class=select>` the same way.

templates/articles/proposal.html
vinyll: Tests are done. Next step may be merging.

PR #65, opened 2017-11-16
Footer

johangiraud:


static/css/styles.css
```diff
@@ -735,4 +737,26 @@ button.menu-button + nav a {
...
+    color: #fff;
+    position: fixed;
+    bottom: 0;
+    left: 0;
+    z-index: -1;
+}
+
+footer p {
+    position: absolute;
+    transform: translate(-50%,-50%);
```
vinyll: missing spae after `, ` :p

templates/base.html
```diff
@@ -51,7 +51,9 @@
 {% block body %}{% endblock %}
 </main>
 
-
+<footer>
+  <p>© Cafébabel — 2017</p>
+</footer>
 
 <script src=https://rawgit.com/cafebabel/styleguide/master/script.js></script>
```
vinyll: This is also a script that should be ported into this repo

vinyll (approved):

templates/base.html
johangiraud: sure

static/css/styles.css
```diff
@@ -41,8 +41,15 @@ summary {
...
     display: block;
+    width: 100%;
+    background-color: #fff;
+    min-height: 100vh; /* https://stackoverflow.com/questions/1509589/full-viewport-height-scaling-div-just-css-no-js-possible */
+}
+
+#content {
     max-width: 120rem;
-    margin: 0 auto;
+    margin: 0 auto 70rem auto;
```
vinyll, line 51: I wonder how much grid would do this job with `justify-self: center`.

static/css/styles.css
```diff
@@ -680,7 +687,7 @@ button.menu-button + nav a {
 }
 
 @media screen and (min-width: 750px) {
-    .profile-page main {
+    .profile-page main #content {
         display: grid;
```
vinyll, line 704: I guess it would be interesting to use grid not only for larger devices but also small screen and responsiveness.
It feels somehow like we are mixing grids and non-grid styles where it could take its place; but I'm not familiar enough with grid to tell more than a feeling…

vinyll (approved):
I don't want to block this PR and it's all good to me; however as the comments say it feels like we could use grid more appropriately.
It might be interesting to tweak them a little further than the general layout and dive deeper into it.

static/css/styles.css
johangiraud: You are right about mixing grid and non-grid styles but what's the point to use it when it comes to 1 column centered ? I guess it is just to act strict and rigorous, which I ask myself if it worth the shot ? Let's talk about it tonight. I merge the PR, but keep in mind it can be changed later

vinyll: It's more about organizing the page withing the content than strictness. I didn't mean to put a grid just for a centering ;)

johangiraud: ok, I'll try then

PR #76, opened 2017-11-24
default link style

johangiraud:


static/css/styles.css
```diff
@@ -476,28 +476,44 @@ a:active {
     text-decoration: none;
 }
 
-p a {
+a {
     position: relative;
 }
 
-p a:after {
+a:after {
```
vinyll: Are you sure you want this on all existing list of the whole site (including header, nav, footer, …)?

static/css/styles.css
```diff
@@ -476,28 +476,44 @@ a:active {
...
+    left: 0;
+    bottom: 0;
+    width: 0;
+    transition: all .3s;
+}
+a:hover:after {
+    width: 100%;
+}
+
+.article-page article p a:after {
```
vinyll, line 498: Why do you redo what you already have above?

static/css/styles.css
johangiraud: That's the default behavior, then I have to overwrite it if I want something for a specific element

static/css/styles.css
johangiraud: I don't understand what you're saying

vinyll: you're doing here `content: ''` (line 498) that is already default in `a` (line 484).
Same for `position:absolute`, `bottom: 0`, …

johangiraud: oh yes right

vinyll (approved):

PR #78, opened 2017-11-27
Style correction

johangiraud:


static/css/styles.css
```diff
@@ -417,10 +418,10 @@ label[for=remember] {
     border: 0.1rem solid #bde513;
 }
 
-label:hover:before {
+/*.select label:hover:before {
     border : 0.1rem solid #bde513 !important;
     transition: border-color .3s;
-}
+}*/
```
vinyll: You may delete this code if it's now useless

static/js/script.js
```diff
@@ -10,7 +10,13 @@ Array.from(document.querySelectorAll('a')).forEach(a => {
...
-}, 3000);
\ No newline at end of file
+}, 3000);
+
+/* display profile social network field on click */
+const socialIcon = document.querySelector('.social-networks li');
+socialIcon.addEventListener('click', (event) => {
+    event.preventDefault();
+    this.document.querySelector('label').classList.toggle('active');
+});
```
vinyll: Missing empty line at the end of line ;)

static/js/script.js
```diff
@@ -10,7 +10,13 @@ Array.from(document.querySelectorAll('a')).forEach(a => {
...
-
 setTimeout(() => {
     document.querySelector('.flashes').classList.add('hidden');
-}, 3000);
\ No newline at end of file
+}, 3000);
+
+/* display profile social network field on click */
+const socialIcon = document.querySelector('.social-networks li');
+socialIcon.addEventListener('click', (event) => {
```
vinyll: 👍

templates/articles/drafts/edit.html
```diff
@@ -47,6 +51,7 @@
   <div class=textarea>
     <label for=body>Body</label>
     <textarea id=body name=body required>{{ article.body }}</textarea>
+    <pre><b>**bold**</b> <i>_italic_</i> #H1 ##H2 ###H3 link: [link text](https://www.google.com)<br>Image: ![alt text](https://www/path/to/img.jpg "Logo Title")</pre>
```
vinyll: I'd suggest to take another example than Google® 💃

templates/articles/edit.html
```diff
@@ -51,7 +51,7 @@
   <div class=textarea>
     <label for=body>Body</label>
     <textarea id=body name=body required>{{ article.body }}</textarea>
-    <pre><b>**bold**</b> <i>_italic_</i> #H1 ##H2 ###H3 link: [link text](https://www.google.com) ![alt text](https://github.com/adam-p/markdown-here/raw/master/src/common/images/icon48.png "Logo Title")</pre>
+    <pre><b>**bold**</b> <i>_italic_</i> #H1 ##H2 ###H3 link: [link text](https://www.google.com)<br>Image: ![alt text](https://www/path/to/img.jpg "Logo Title")</pre>
```
vinyll: This code is repeated multiple times. It might be time for a commont layout / a include / a macro

templates/auth/login.html
```diff
@@ -2,8 +2,8 @@
 {% from "_forms.html" import render_field_with_errors, render_field %}
 
 {% block body %}
-<h1>{{ _('Login') }}</h1>
 <form action="{{ url_for_security('login') }}" method="POST" name="login_user_form">
+  <h1>{{ _('Login') }}</h1>
```
vinyll: Just being curious, why is it better within the `<h1>`?

templates/base.html
```diff
@@ -50,6 +46,9 @@
 {% include "security/_messages.html" %}
 
 <main>
+  <a href={{ url_for('home') }} class=logo>
+    <img src=https://rawgit.com/cafebabel/styleguide/master/img/logo-cafebabel.svg alt="logo Cafébabel">
```
vinyll: You probably want to import this into the project

templates/profile.html
```diff
@@ -11,13 +11,49 @@ <h1>Create a profile</h1>
...
-          <li><a data-edit-name=socials-facebook href="{{ user.profile.socials.facebook }}" class=icon-facebook></a></li>
-          <li><a data-edit-name=socials-instagram href="{{ user.profile.socials.instagram }}" class=icon-instagram></a></li>
-          <li><a data-edit-name=socials-medium href="{{ user.profile.socials.medium }}" class=icon-medium></a></li>
-          <li><a data-edit-name=socials-flickr href="{{ user.profile.socials.flickr }}" class=icon-flickr2></a></li>
-          <li><a data-edit-name=socials-pinterest href="{{ user.profile.socials.pinterest }}" class=icon-pinterest2></a></li>
+      <ul class=social-networks>
+          <li>
+              <a data-edit-name=socials-twitter href="{{ user.profile.socials.twitter }}" class=icon-twitter></a>
+              <label for=social-twitter>
+                  <span class=icon-twitter></span>
```
vinyll, line 18: Why do we have `<a class=icon-twitter>` followed by `<span class=icon-twitter></span>`?

static/js/script.js
johangiraud: oh

templates/articles/drafts/edit.html
johangiraud: haha sure

templates/articles/edit.html
johangiraud: That's what I thought.. Afterwards ;-) Ok I imported it now.

templates/auth/login.html
johangiraud: Was already within a `<h1>`. Makes sense since it's the page title, doesn't it ?

templates/base.html
johangiraud: ah oui !

templates/profile.html
johangiraud: Cause I want twice the icon. I'm sure it's not the best way to do it, but I had no other idea yet

static/js/script.js
```diff
@@ -20,3 +20,15 @@ socialIcon.addEventListener('click', (event) => {
     event.preventDefault();
     this.document.querySelector('label').classList.toggle('active');
 });
+
```
johangiraud: Here I would like to do this but in vanilla js (pure js) : 
```
$('input').focus(function(){
  $(this).parent().addClass('active completed');
});

$('input').focusout(function(){
  if($(this).val() === '')
    $(this).parent().removeClass('completed');
  $(this).parent().removeClass('active');
})
```
Any idea on how I could get this working ? @vinyll

vinyll: If it's only one input element:
```js
const input = document.querySelector('input')
input.addEventListener('focus', event => {
  event.target.parent.classList.add('active', 'complete')
})

input.addEventListener('blur' => {
  if(event.target.value === '') {
    event.target.classList.remove('completed', 'active')
})
```

templates/profile.html
vinyll: There should be other ways indeed but if you want it twice it should do for now

templates/auth/login.html
vinyll: This would semantically be the form's title then

static/js/script.js
johangiraud: you don't need any semicolon at the end of every line then ? and for `'blur' => {` don't you want need : `'blur', (event) => {` ?

templates/auth/login.html
johangiraud: ah ok, well login is also the form's name. same same

static/js/script.js
johangiraud: i used : https://codepen.io/legeoffrey/pen/VYMLNq?q=input&limit=all&type=type-pens as an example. And I want to use it for cfbb login form

templates/base.html
```diff
@@ -57,5 +57,8 @@
 <footer>
   <p>© Cafébabel — 2017</p>
 </footer>
+<div class=preloader>
```
vinyll: I wonder if it wouldn't make sense to use an `id` here instead of `class`. What's your opinion?

johangiraud: I never know when it's best to use ID's. I always think _qui peut le plus peut le moins_ ! a class can appear in severals places when an id can't.

vinyll: That's correct. It seems like the loader is covering the entire page and therefore cannot have multiple on a page. That's what made me wonder about this; making it obvious that it can only be unique. Just a thought.

vinyll (approved):

templates/base.html
johangiraud: ok !

static/js/script.js
```diff
@@ -1,17 +1,17 @@
-/* Preloader fadeOut */
+/* Preloader fadeout when page is loaded */
 window.onload = function() {
-    const preloader = document.getElementsByClassName('preloader')[0];
-    preloader.className += ' fadeout';
+    const preloader = document.querySelector('#preloader');
+    preloader.classList.add('fadeout');
     setTimeout(function(){
```
vinyll: Use the arrow function for consistency with the rest of the code:

```js
setTimeout(() => {
  …
```

static/js/script.js
```diff
@@ -1,17 +1,17 @@
...
         preloader.style.display = 'none';
```
vinyll: You might want to use a class instead (like _hidden_ for example) which is more performant than generating styles in JS, and play with `classList` as you do in other places.

static/js/script.js
vinyll: Here's a Vanilla equivalent: https://codepen.io/anon/pen/xPyVVY/

templates/base.html
```diff
@@ -43,6 +43,10 @@
   </ul>
 </nav>
 
+<nav>
```
vinyll: Delete

static/js/script.js
johangiraud: ok !

static/js/script.js
johangiraud: sure

static/js/script.js
johangiraud: and for this one : `window.onload = function() {` should I also use : `window.onload = () => {` or it's different ?

vinyll: In recent JavaScript, you should always use arrow functions (`=> {`).
There are a very few specific exception but they are so seldom you may ignore them ;)

static/js/script.js
```diff
@@ -19,30 +17,36 @@ Array.from(document.querySelectorAll('a')).forEach(a => {
...
 }
 
 /* display profile social network field on click */
-const socialIcon = document.querySelector('.social-networks li');
-socialIcon.addEventListener('click', (event) => {
-    event.preventDefault();
-    this.document.querySelector('label').classList.toggle('active')
-});
+const socialIcon = document.querySelector('.social-networks li')
+Array.from(document.socialIcon.forEach(li => {
```
vinyll: because you have used parenthesis everywhere in your script even for single or no parameter, you should also write `(li) => {` here

static/js/script.js
```diff
@@ -19,30 +17,36 @@ Array.from(document.querySelectorAll('a')).forEach(a => {
...
+        })
+    }
+}
 
 /* animation login fields */
 Array.from(document.querySelectorAll('form[name=login_user_form] input')).forEach(input => {
     const parent = input.parentElement
-    input.addEventListener('focus', _ => {
-    console.log('focus')
+    console.log('hey')
```
vinyll: remove me

vinyll (changes requested):

static/js/script.js
```diff
@@ -26,19 +26,17 @@ if (flashes) {
...
-    if (socialIcon) {
-        socialIcon.addEventListener('click', (event) => {
-            event.preventDefault()
-        this.document.querySelector('label').classList.toggle('active')
-        })
-    }
+if (socialIcon) {
+    socialIcon.addEventListener('click', (event) => {
+        event.preventDefault()
+    this.document.querySelector('label').classList.toggle('active')
```
vinyll: `this` is useless here. `docoument` is global.

vinyll (changes requested):

static/css/styles.css
```diff
@@ -358,24 +352,30 @@ form[name=login_user_form] ul {
...
     margin-left: -1rem;
 }
-form[name=login_user_form]:before,
-form[name=login_user_form]:hover:before {
-    border-color: rgba(0, 40, 60, 0);
-    border-bottom-color: #00283C;
+form[name=login_user_form] ul:before,
+form[name=login_user_form] ul:hover:before {
+    border-color: rgba(255, 0, 0, 0);
+    border-bottom-color: red;
```
vinyll: is this for debugging or is `red` the actual color your want?

johangiraud: red is the color I want

static/js/script.js
johangiraud: even if I want `.active` class to be added to the `label` inside the clicked `li` ?

static/js/script.js
johangiraud: ok

static/css/styles.css
```diff
@@ -352,28 +357,43 @@ form[name=login_user_form] ul {
...
+form[name=change_password_form] ul li,
+form[name=register_user_form] ul li,
+form[name=forgot_password_form] ul li {
     text-align: center;
     font-size: 1.2rem;
     color: red;
 }
 
-form[name=login_user_form] ul:after, form[name=login_user_form] ul:before {
+form[name=login_user_form] ul:after, form[name=login_user_form] ul:before,
```
vinyll: new line after comma for having each selector on its line like [you did on the `:hover`](https://github.com/cafebabel/cafebabel.com/pull/78/files/edc0ab8265da5807c9bfc11bcf32cabb66cff046..df19f95d22a43b5ff7e1c16fe6377f330346591c#diff-7962f8c018310ce9bd37011f4b6ce5e3R390); easier to read in the longer term

static/js/script.js
```diff
@@ -27,9 +27,10 @@ if (flashes) {
 /* display profile social network field on click */
 const socialIcon = document.querySelector('.social-networks li')
 if (socialIcon) {
-    Array.from(document.querySelector('.social-networks li')).forEach((li) => {
-            socialIcon.addEventListener('click', (event) => {
-                event.preventDefault()
+    Array.from(socialIcon).forEach((li) => {
+        console.log('oink')
```
vinyll: remove log

static/js/script.js
```diff
@@ -27,9 +27,10 @@ if (flashes) {
...
+        li.addEventListener('click', (event) => {
```
vinyll: If it's clickable, it should be a `button` or a `a`, else this is just fully breaking any accessibility for sure.

vinyll (changes requested):

static/css/styles.css
johangiraud: ok

static/js/script.js
johangiraud: oui

static/js/script.js
```diff
@@ -25,10 +25,9 @@ if (flashes) {
 }
 
 /* display profile social network field on click */
-const socialIcon = document.querySelector('.social-networks li')
+const socialIcon = document.querySelector('.social-networks li a')
 if (socialIcon) {
     Array.from(socialIcon).forEach((li) => {
```
vinyll: rename `li` for consistency

vinyll (changes requested):

static/js/script.js
johangiraud: ok

static/js/script.js
```diff
@@ -25,27 +25,29 @@ if (flashes) {
...
+            socialIcon.classList.add('active')
         })
     })
 }
 
 /* animation login fields */
 Array.from(document.querySelectorAll('form[name=login_user_form] > div > input')).forEach(input => {
     const parent = input.parentElement
-    input.addEventListener('focus', () => {
+    console.log(parent)
```
vinyll: remove `console.log`

templates/profile.html
```diff
@@ -16,42 +16,42 @@ <h2>About</h2>
...
+                  <input type=url id=socials-facebook name=socials-facebook value="{{ user.profile.socials.facebook }}">
               </label>
           </li>
           <li>
               <a data-edit-name=socials-instagram href="{{ user.profile.socials.instagram }}" class=icon-instagram></a>
-              <label for=social-instagram>
+              <label for=socials-instagram>
                   <span class=icon-instagram></span>
-                  <input type=url id=social-instagram>
+                  <input type=url id=socials-instagram>
```
vinyll: missing `name` and `value`.
Same for fields below

templates/auth/login.html
```diff
@@ -3,7 +3,7 @@
 
 {% block body %}
 <h1>{{ _('Login') }}</h1>
-<form action="{{ url_for_security('login') }}" method="POST" name="login_user_form">
+<form action="{{ url_for_security('login') }}" method="POST" name="login_user_form" class=authentication-form>
```
vinyll, line 6: If you're using this class, you may want to adapt your CSS accordingly

templates/profile.html
```diff
@@ -102,16 +85,17 @@ <h3>{{ article.title }}</h3>
...
-        <time>Published on October 11th 2017</time>
-        <h3>Lisbon: Nomad's land</h3>
-    </article>
-    <article>
-        <img src="http://m.cafebabel.com/cache/27/f3/27f3816aed817cb844a51012de66c8fa.jpg?cache=1280x500" alt="Lisbon digital nomad cover">
-        <time>Published on October 11th 2017</time>
-        <h3>Lisbon: Nomad's land</h3>
-    </article>
+      {% for article in articles %}
+      <article>
```
vinyll: remove indentation. You're at the same level as `<h2>`

templates/profile.html
```diff
@@ -102,16 +85,17 @@ <h3>{{ article.title }}</h3>
...
+          <a href={{ url_for('articles.detail', article_id=article.id, slug=article.slug) }}>
+              {% if article.has_image %}
+                <img src={{ article.image_url }}>
```
vinyll: Missing `alt`

vinyll (changes requested):

static/js/script.js
johangiraud: ok

templates/auth/login.html
johangiraud: yep

templates/profile.html
johangiraud: ok

templates/profile.html
johangiraud: yes

vinyll (approved):

PR #100, opened 2017-12-06
Editing place style

johangiraud:


cafebabel/static/css/styles.css
```diff
@@ -767,6 +768,62 @@ button.menu-button + nav a {
...
+.create-article-page canvas.active {
+    opacity: .8;
+}
+
+.create-article-page .file label {
+    position: absolute;
+    top: 50%;
+    left: 50%;
+    text-align: center;
+    transform: translate(-50%,-50%);
```
vinyll: missing space after comma

cafebabel/static/js/script.js
```diff
@@ -56,3 +56,56 @@ Array.from(document.querySelectorAll('.authentication-form > div > input')).forE
...
+    })
+    addListenerMulti(fileInput, 'dragleave blur drop', () => {
+        dropArea.classList.remove('active')
+    })
+    fileInput.addEventListener('change', () => {
+        document.querySelector('.file label').innerHTML(' ')
+    })
+}
+
+/* tv noise effect https://codepen.io/eugene-bulkin/pen/zEgyH how to know how "heavy" is it? */
```
vinyll: isolate this script into a different file?

cafebabel/static/js/script.js
```diff
@@ -56,3 +56,56 @@ Array.from(document.querySelectorAll('.authentication-form > div > input')).forE
...
+    (function loop() {
+        toggle = !toggle
+        if (toggle) {
+            requestAnimationFrame(loop)
+            return
+        }
+        noise(context)
+        requestAnimationFrame(loop)
+    })();
+}
```
vinyll: Missing empty line

cafebabel/templates/articles/drafts/create.html
```diff
@@ -1,11 +1,17 @@
 {% extends "base.html" %}
 
+{% block body_class %}create-article-page{% endblock %}
 {% block body %}
   <form method=post action={{ url_for('drafts.create') }} enctype=multipart/form-data>
-  <div class=text>
-    <label for=title>Title</label>
-    <input id=title name=title required>
+  <canvas></canvas>
```
vinyll: We might want to generate this tag dynamically somehow.
Eventually we'll see for that later

cafebabel/static/js/script.js
```diff
@@ -56,3 +56,56 @@ Array.from(document.querySelectorAll('.authentication-form > div > input')).forE
...
+const dropArea = document.querySelector('canvas')
+if (fileInput) {
+    addListenerMulti(fileInput, 'dragenter focus click', () => {
+        dropArea.classList.add('active')
+    })
+    addListenerMulti(fileInput, 'dragleave blur drop', () => {
+        dropArea.classList.remove('active')
+    })
+    fileInput.addEventListener('change', () => {
+        document.querySelector('.file label').innerHTML(' ')
```
vinyll, line 98: Why removing the label when the file has changed?

cafebabel/static/js/script.js
```diff
@@ -56,3 +56,56 @@ Array.from(document.querySelectorAll('.authentication-form > div > input')).forE
...
+    const context = canvas.getContext('2d')
+
+    function resize() {
+        canvas.width = window.innerWidth
+        canvas.height = window.innerHeight
+    }
+    resize();
+    window.onresize = resize
+
+    function noise(context) {
```
vinyll: Shouldn't we move this function into its own file? (eg: _noise.js_)

vinyll (commented):
I guess we could imagine something generic like a class that would make the tv-blur effect.
This should probably get refactored as some code is very specific and might make troubles if attempting to change the effect in the future.
I can get a hand onto it, there's a couple of small things to change in the meantime

cafebabel/static/js/script.js
johangiraud: Im not done with that, I would like to display the file name as soon as it's dropped (or chosen). I was thinking of displaying it replacing the label content. But maybe not the best from an accessibility point of view ?

cafebabel/static/js/script.js
johangiraud: ok, why ? to get rid of the useless weight when other pages are visited <?

cafebabel/static/js/script.js
johangiraud: yep, cause every time the window size is change we call this function ?

cafebabel/static/js/script.js
johangiraud: ok

cafebabel/templates/articles/drafts/create.html
johangiraud: what's the problem with having `<canvas>` tag inside the template ?

cafebabel/static/css/styles.css
```diff
@@ -157,12 +157,18 @@ main {
...
+.icon-file:before {
+    content: '\e927';
+}
 .icon-profile:before {
     content: '\e902';
 }
 .icon-checkmark:before {
     content: '\ea10';
 }
+.icon-heart:before {
```
vinyll, line 169: The name "heart" shows what is looks like instead of what it does.
If you ever change the icon, you will need to change the name also.

As an example you gave the class "share-button" for a button to share. I'm not sure what it looks like, but I know what it does. If it looks like that triangle aside, you could have called it ".aside-triangle-button"; we would have understood what it looks like but not what it does.
Another example is ".menu-button" that you gave the name to what it does, you didn't call it ".three-lines-icon".

Is the ".icon-heart" for favoriting?

A best practice is to tell what it does as adapting the design won't make you change the HTML, but the CSS ;)

cafebabel/static/css/styles.css
```diff
@@ -157,12 +157,18 @@ main {
...
+    content: '\e9da';
+}
 .icon-arrow-down:before {
```
vinyll, line 172: Same here. I'm sure it looks like an arrow pointing down, but what is it for? Is it the select button arrow?

cafebabel/static/css/styles.css
```diff
@@ -750,24 +766,65 @@ button.menu-button + nav ul li a:after {
...
+button.menu.active + nav ul li {
     padding-bottom: 0;
 }
 
-button.menu-button + nav a {
+button.menu + nav a {
     display: block;
 }
 
+button.share-button:before {
```
vinyll: I wonder why you chose to have `button.share-button` instead of `button.share`?

cafebabel/static/css/styles.css
```diff
@@ -809,16 +873,42 @@ button.menu-button + nav a {
     opacity: 0;
 }
 
+.create-article-page h1.text {
```
vinyll: I'm confused with the selector at line 880.
`h1` can have the `.text` class but it can also be on a wrapper that hold form items?

vinyll: Back to this. Why is it called "text"? If it's supposed to be plain text, why not using a `<p>`?

cafebabel/templates/articles/drafts/create.html
vinyll: `<canvas>` right there has no content therefore no meaning.
It's an empty tag and reading the HTML, it's obvious we can remove it.

Also `<canvas>` could be used for other purposes in the project. But if we ever attempt to use it again, it will be decorated but the noise effect. Maybe we should at least apply a class on the canvas like `.noise` that we could use in the JS.

cafebabel/templates/articles/drafts/create.html
```diff
@@ -5,50 +5,59 @@
...
+        <select id=language name=language required>
+          <optgroup label=Language>
+          {% for language in get_languages() %}
+          <option value={{language[0]}}>{{ language[1] }}</option>
+          {% endfor %}
+        </select>
+      </label>
+    </div>
+    <div class=select>
+      <label for=category>
```
vinyll, line 28: Even though we are not making the site highly accessible, I think removing text from the `<label>` will completely remove any hope not to get too far away from it.

Could we eventually wrap the real label into a `<span>` and hide it from the screen in css? (which should still make it browsable for non screen displays). Suggestions or documentation are welcome if you have further info on this.

cafebabel/templates/articles/drafts/create.html
johangiraud: yep

cafebabel/static/css/styles.css
johangiraud: ok, I changed them to `.language`, `.follow` and `.main-menu`

cafebabel/static/css/styles.css
johangiraud: I switched it to `h1.edit`, best

cafebabel/templates/articles/drafts/create.html
johangiraud: What do you mean bu "wrapping the real label into a `<span>`"? like : 
```
<span>
<label for=category>
<select...
</label>
</span>
```
weirdo, isn't it ?

.idea/uiDesigner.xml
```diff
@@ -0,0 +1,124 @@
...
+      </item>
+      <item class="javax.swing.JToolBar$Separator" icon="/com/intellij/uiDesigner/icons/toolbarSeparator.png" removable="false" auto-create-binding="false" can-attach-label="false">
+        <default-constraints vsize-policy="0" hsize-policy="0" anchor="0" fill="1" />
+      </item>
+      <item class="javax.swing.JScrollBar" icon="/com/intellij/uiDesigner/icons/scrollbar.png" removable="false" auto-create-binding="true" can-attach-label="false">
+        <default-constraints vsize-policy="6" hsize-policy="0" anchor="0" fill="2" />
+      </item>
+    </group>
+  </component>
+</project>
```
vinyll, line 124: missing empty line at the end of file

cafebabel/static/css/styles.css
```diff
@@ -903,10 +915,26 @@ button.share-button.active:after {
...
     max-width: 75rem;
     margin: auto;
+    text-align: justify;
 }
 
-.create-article-page .select {
-    display: inline-block;
+.create-article-page .article-settings:after { /* https://codepen.io/gcyrillus/pen/ruhCg */
+    content: '';
+    padding-right: 100%;
```
vinyll: does this work without a `display: block (or inline-block)`?

cafebabel/static/js/script.js
```diff
@@ -67,7 +67,7 @@ Array.from(document.querySelectorAll('.authentication-form > div > input')).forE
 })
 
 /* animation title field create article */
-Array.from(document.querySelectorAll('h1.text input')).forEach(inputh1 => {
+Array.from(document.querySelectorAll('h1.edit input')).forEach(inputh1 => {
```
vinyll: I think you've use parenthesis also for single argument in your other scripts?

cafebabel/templates/articles/drafts/edit.html
```diff
@@ -1,49 +1,71 @@
...
 
+{% block body_class %}create-article-page{% endblock %}
 {% block body %}
 <form method=post action={{ url_for('drafts.edit', draft_id=article.id) }} enctype=multipart/form-data>
   <input type=hidden name=id value="{{ article.id }}">
-  <div class=text>
+  <div class=file>
+      <label for=image>Image</label>
+      {% if article.has_image %}
+          <img src={{ article_image_url(article) }} alt>
```
vinyll: Should we put a a default `alt` value? Is this W3C valid?

cafebabel/templates/articles/drafts/edit.html
```diff
@@ -1,49 +1,71 @@
...
+          <div class=checkbox>
+            <label for=delete-image>Delete image</label>
+            <input type=checkbox name=delete-image id=delete-image value=1>
+          </div>
+      {% endif %}
+      <input type=file id=image name=image>
+  </div>
+  <h1 class='edit active'>
```
vinyll: Please keep the double quote convention here too

cafebabel/templates/articles/drafts/edit.html
```diff
@@ -1,49 +1,71 @@
...
+            <optgroup label=Language>
+              {% for language in get_languages() %}
+              <option value={{language[0]}}{% if article.language == language[0] %} selected{% endif %}>{{ language[1] }}</option>
+              {% endfor %}
+            </optgroup>
+          </select>
+        </label>
+      </div>
+      <div class=select>
+        <label for=category>
```
vinyll, line 35: My concern was that we do have a label tag, but no label content.
Suggestion was to insert a text that would be eventually removed from the screen with CSS `<label> <span>My non visible label</span>…`.

However I'm not aware whether or not `<optgroup label=xxx>` is accessible and enough for replacing a label? I can find documentation on regular label but not optgroup poorly documented and not so widely used.

In short: I don't have enough knowledge on this, so whatever you know/feel is best for the project 🥉

cafebabel/templates/articles/edit.html
```diff
@@ -63,6 +14,66 @@
     </div>
     {% endif %}
     <input type=file id=image name=image>
+  </div>
+    <h1 class='edit active'>
```
vinyll: double quotes in HTML

.idea/uiDesigner.xml
johangiraud: dunno why this file stand here :/

cafebabel/static/css/styles.css
johangiraud: no

cafebabel/templates/articles/drafts/edit.html
johangiraud: Yes, we should, I haven't touched it, i guess you places it like that. Somewhere else, you've written something like `alt="image for {{ article.title }}"`. I think it's better.

cafebabel/templates/articles/drafts/edit.html
johangiraud: ok

cafebabel/templates/articles/drafts/edit.html
johangiraud: haha ok

cafebabel/templates/articles/drafts/create.html
vinyll: What I meant was to have `<label><span>My Label Title</span>…</label>` so that the span could be manipulated in CSS

johangiraud: ok sure, ensuite `display:none;` sur le `span`, carrément ?

.idea/uiDesigner.xml
vinyll: We ca just get rid of this `.idea` folder

cafebabel/static/css/styles.css
```diff
@@ -1160,14 +1161,18 @@ button.follow.active:after {
...
+}
+
 .pitch-page .select label:after {
-    top: 9rem;
+    top: 5rem;
 }
 
 @media screen and (min-width: 750px) {
-    .pitch-page .select label:after {
+    /*.pitch-page .select label:after {
```
vinyll: delete me

cafebabel/templates/articles/to-translate.html
```diff
@@ -40,15 +40,15 @@ <h1>
           {% endif %}
           <time>Published on {{ article.creation_date.date() }}</time>
           <h2>{{ article.title }}</h2>
-          <ul>
+
             {% if article.is_translation %}
               {% if article.original_article.language != to_language_code %}
-                <li><a href={{ url_for('translations.create') }}?lang={{ to_language_code }}&original={{ article.original_article.id }}>Translate in {{ to_language_label }}</a></li>
+                <a href={{ url_for('translations.create') }}?lang={{ to_language_code }}&original={{ article.original_article.id }}><button>Translate in {{ to_language_label }}</button></a>
```
vinyll: `<a><button>`??

cafebabel/templates/articles/to-translate.html
```diff
@@ -40,15 +40,15 @@ <h1>
...
               {% endif %}
             {% else %}
-              <li><a href={{ url_for('translations.create') }}?lang={{ to_language_code }}&original={{ article.id }}>Translate in {{ to_language_label }}</a></li>
+              <a href={{ url_for('translations.create') }}?lang={{ to_language_code }}&original={{ article.id }}><button>Translate in {{ to_language_label }}</button></a>
```
vinyll: same here

vinyll (changes requested):

PR #101, opened 2017-12-07
Update profile page should be split into edit and detail

vinyll:
- [x] Social networks tooltips should close when clicking outside.
- [x] Upload/delete a profile picture


cafebabel/core/mixins.py
```diff
@@ -0,0 +1,48 @@
+from flask import current_app
```
davidbgk: Useless import?

cafebabel/core/mixins.py
```diff
@@ -0,0 +1,48 @@
...
+from mongoengine import signals
```
davidbgk: Same here.

cafebabel/core/mixins.py
```diff
@@ -0,0 +1,48 @@
...
+
+from .. import db
+
+
+class UploadableImageMixin:
+    has_image = db.BooleanField(default=False)
```
davidbgk, line 5: Not fond of having fields declared in many places, I would keep that one on `Article` and thus get rid of the `db` dependency in the mixins.

davidbgk (changes requested):

cafebabel/core/mixins.py
vinyll: It' about to be duplicated for User that will have the same field. that's why the split into a mixin.
Plus the Mixin is explicitely dependant of that field.

davidbgk: Fair enough, I wasn't aware of the future use in User 👍

davidbgk (approved):

cafebabel/templates/base.html
```diff
@@ -15,7 +15,7 @@
 {% if current_user.is_authenticated %}
 <nav>
   <ul>
-    <li><a href={{ url_for('users.profile') }}>Profile</a></li>
+    <li><a href={{ url_for('users.me') }}>Profile</a></li>
```
johangiraud: Maybe `my-profile` instead of `me` would be more relevant ?

cafebabel/templates/articles/_card.html
```diff
@@ -0,0 +1,9 @@
+<article>
+  <h2>{{ article.title }}</h2>
+  {% if article.has_image %}
+  <img src="{{ article.image_url }}" alt="Image for {{ article.title }}">
+  {% endif %}
+  <p>
+    {{ article.summary }}
+  </p>
+</article>
```
johangiraud: why is it called `_card` ?

cafebabel/templates/users/macros.html
```diff
@@ -0,0 +1,9 @@
+{% macro social_field(name) %}
+<li>
+  <a href=# class=icon-{{ name }}></a>
+  <label for=social-{{ name }}>
+    <span class=icon-{{ name }}></span>
+    <input type=url id=social-{{ name }} name=socials-{{ name }} value="{{ user.profile.socials.get(name) }}">
+  </label>
+</li>
+{% endmacro %}
```
johangiraud: same : why is it called `macros` ? For me it doesn't say enough what it is about

cafebabel/templates/users/detail.html
```diff
@@ -0,0 +1,44 @@
...
+<section>
+  <h2>About</h2>
+  {% if user.is_me() or current_user.has_role('editor') %}
+  <a href={{ url_for('users.edit', id=user.id) }}>Edit this profile</a>
+  {% endif %}
+  {% if user.profile.has_image %}
+  <img src={{ user.profile.image_url }} alt="Picture of {{ user }}">
+  {% endif %}
+
+  <ul class=social-networks>
```
johangiraud: Here we need to display icons only if the user previously added his social details. So we need a `if` statement.

cafebabel/templates/users/macros.html
vinyll: I was thinking of storing all macros that we may have for the users module. Just like we have users/views for the views and users/models for the models, each having the users related functionalities.
For now we only have the `social_field` macro though.
How would you see this?

cafebabel/templates/users/detail.html
vinyll: We do have `if` for each individual social field below.
A user may insert some social and not all, therefore we still need the `ul`.

cafebabel/templates/base.html
vinyll: We cannot insert hyphens here as it relates to the name of a function in Python (which don't permit hyphens)

vinyll:
@davidbgk: this PR is complete. New review is welcome :)

cafebabel/templates/base.html
johangiraud: yea well `myprofile` then

cafebabel/templates/users/detail.html
johangiraud: if the user doesn't fill any of these `li`, then we get an empty `ul`? > not good

cafebabel/templates/users/macros.html
johangiraud: ok makes more sense as long as there are severals. But still I would pick a name that shows what macro it is like `profile-marcos`?

vinyll: Don't you feel like it would be redundant to have `users/profile-macros.html`?
When we'll have some for the article, does it mean we'd call them `articles/articles-macros.html`?

johangiraud: we'll have `input id=social-{{ name }}..` only on profile since we can register the social only here. That's why i suggested `profile-macros`

vinyll: indeed, that's why I put them in the `users/` folder ;)
Anyways, I'll call it _users/profile-macros.html_ if it makes more sense to you 🤝

vinyll:
Tests are fixed in https://github.com/cafebabel/cafebabel.com/pull/111

cafebabel/templates/users/macros.html
johangiraud: cool !

vinyll:
Seems good for you, I'm merging 👀

PR #112, opened 2017-12-20
anonymous user is redirected to login when creating a draft

vinyll:
the 2 broken tests are fixed in https://github.com/cafebabel/cafebabel.com/pull/111

### DIFF
--- a/cafebabel/articles/drafts/views.py
+++ b/cafebabel/articles/drafts/views.py
@@ -17,8 +17,8 @@ def list():
 
 
 @drafts.route('/new/', methods=['get', 'post'])
-@editor_required
 @login_required
+@editor_required
 def create():
     if request.method == 'POST':
         article = Article().save_from_request(request)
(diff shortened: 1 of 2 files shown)


cafebabel/articles/drafts/views.py
```diff
@@ -17,8 +17,8 @@ def list():
 
 
 @drafts.route('/new/', methods=['get', 'post'])
-@editor_required
 @login_required
+@editor_required
```
davidbgk, line 21: Not fond of the fact that order matters here, maybe add automagically `login_required` when `editor_required` is added? It looks more consistent to me.

vinyll: What is not found?
I'm ok to have `login_required` embedded into `editor_required` in a different PR for keeping its original purpose on this one.

davidbgk: Without the U? https://www.vocabulary.com/dictionary/fond

OK to put it in a subsequent PR

davidbgk (approved):

cafebabel/articles/drafts/views.py
vinyll: Issue [opened here](https://github.com/cafebabel/cafebabel.com/issues/114) for memory.

PR #116, opened 2018-01-04
Manually dealing with unique translation per language and article

vinyll:


cafebabel/articles/translations/models.py
```diff
@@ -38,8 +37,19 @@ def clean(self):
         except Article.DoesNotExist:
             pass
 
+    @classmethod
+    def check_duplicate(cls, sender, document, **kwargs):
+        if document.id:
```
davidbgk: Thinking again about it, in case of an update we have to make the check too like `Translation.objects(original_article=document.original_article, language=document.language).count() == 1` or something more clever…

vinyll: You mean if the language was changed during the update?

davidbgk: Yes, proposition:

```python
    @classmethod
    def check_duplicate(cls, sender, document, **kwargs):
        first = Translation.objects(original_article=document.original_article, 
                                    language=document.language).first()
        if (document.id and document.language != first.language) or first:
            raise errors.NotUniqueError(
                'A translation refering to the same article with this '
                'language already exists.')
```

It requires a dedicated test too.

cafebabel/templates/articles/detail.html
```diff
@@ -24,14 +24,16 @@ <h1>{{ article.title }}</h1>
...
             {% endif %}
-            {% for language in get_languages() %}
-                {% if article.language != language[0] and language[0] not in translations_langs %}
+            {% for lang_code, lang_label in get_languages() %}
+                {% if article.language != lang_code and lang_code not in translations_langs %}
                     {% if article.is_translation %}
-                        {% if language[0] != article.original_article.language %}
-                            <li class=to-translate-languages><a href={{ url_for('translations.create') }}?lang={{language[0]}}&original={{ article.original_article.id }}>{{ language[1] }}</a></li>
+                        {% if lang_code != article.original_article.language %}
+                            <li class=to-translate-languages><a href={{ url_for('translations.create') }}?lang={{lang_code}}&original={{ article.original_article.id }}>{{ lang_label }}</a></li>
```
davidbgk: I know it was already the case but let's try to be consistent and add spaces around vars in templates: `{{ lang_code }}` (same below)

cafebabel/articles/models.py
```diff
@@ -63,9 +63,15 @@ def is_translated_in(self, language):
         given that it performs one query per language.
         TODO: turn into a cacheable list of translations when necessary.
         """
+        return bool(self.get_translation(language))
```
davidbgk: I think forcing the `get` to only check the existence is a problem in terms of performances given that it actually retrieves the object for nothing.

cafebabel/articles/models.py
```diff
@@ -23,10 +23,20 @@ class Article(db.Document, UploadableImageMixin):
...
+
     meta = {
         'allow_inheritance': True
     }
 
+    @property
+    def translations(self):
+        if not self._translations:
+            from .translations.models import Translation  # NOQA: circular :/
+            self._translations = list(
```
davidbgk: 1. What about generating a dict `{'lang': article}` once and for all?
2. How do we invalidate that cache in case of newly added translation?

davidbgk (approved):

pointbar:
J'ai testé la fonctionnalité et ouvert une amélioration - #145

PR #118, opened 2018-01-04
editor can add/remove editor role to a profile

vinyll:


cafebabel/templates/users/edit.html
```diff
@@ -43,6 +43,12 @@ <h2>About</h2>
       <label for=about>Bio</label>
       <textarea id=about name=about>{{ user.profile.about }}</textarea>
     </div>
+    {% if current_user.has_role('editor') %}
+    <div class=checkbox>
+      <label for=editor>Editor role</label>
+      <input type=checkbox id=editor name=editor value=1 {% if user.has_role('editor') %}checked{% endif %}>
```
davidbgk, line 49: Always `value=1`?

vinyll: Yes, that's the value of the checkbox.
If empty, value equals to 'on'. Here we have value equals to '' (unchecked) or '1' (checked).
the fact of being checked or not is managed by the `checked` attribute.
 
Or am I missing something with the funky way that checkboxes are managed?

davidbgk (approved):

PR #120, opened 2018-01-04
Avoid config reference in templates

davidbgk:
And only add pertinent emails to the dedicated view.

### DIFF
--- a/cafebabel/templates/articles/proposals/create.html
+++ b/cafebabel/templates/articles/proposals/create.html
@@ -57,13 +57,13 @@
 
     <button type=submit>Suggest this article</button>
   </form>
-  <a class="button proposal-email" href=m&#x61;ilto:{{config.EDITORS_EMAIL_DEFAULT|urlencode|replace('.', '&#46;')|replace('a', '&#x61;')|safe}}>Or send your proposal by email</a>
+  <a class="button proposal-email" href=m&#x61;ilto:{{ EDITORS_EMAIL_DEFAULT|urlencode|replace('.', '&#46;')|replace('a', '&#x61;')|safe }}>Or send your proposal by email</a>
   <script>
-     const editorEmail = {{ editor_emails|to_json }}
+     const editorEmails = {{ EDITOR_EMAILS|to_json }}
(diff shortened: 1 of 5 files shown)


cafebabel/templates/articles/proposals/create.html
```diff
@@ -57,13 +57,13 @@ <h3>Wanna be part of Cafébabel project? That’s the right place</h3>
 
     <button type=submit>Suggest this article</button>
   </form>
-  <a class="button proposal-email" href=m&#x61;ilto:{{config.EDITORS_EMAIL_DEFAULT|urlencode|replace('.', '&#46;')|replace('a', '&#x61;')|safe}}>Or send your proposal by email</a>
+  <a class="button proposal-email" href=m&#x61;ilto:{{ EDITORS_EMAIL_DEFAULT|urlencode|replace('.', '&#46;')|replace('a', '&#x61;')|safe }}>Or send your proposal by email</a>
   <script>
-     const editorEmail = {{ editor_emails|to_json }}
+     const editorEmails = {{ EDITOR_EMAILS|to_json }}
```
vinyll: I was thinking that we may want to encrypt them here in their HTML version like at line 60

cafebabel/tests/test_proposals.py
```diff
@@ -9,6 +9,13 @@ def test_proposal_display_form(app, client):
     assert '<form action=. method=post>' in response
 
 
+def test_proposal_display_emails(app, client):
+    response = client.get('/article/proposal/new/')
+    assert response.status_code == 200
+    assert 'href=m&#x61;ilto:editors%40c&#x61;feb&#x61;bel&#46;com' in response
+    assert '"fr": "redaction@cafebabel.com"' in response
```
vinyll: so that it is not rendered clearly 👀

cafebabel/templates/articles/proposals/create.html
davidbgk: Such a PITA but necessary, thanks 😘

vinyll:
Can't approve as I participated.

_r+_

PR #140, opened 2018-01-10
#US25 - Ajouter des tags à un article

davidbgk:
ETQ éditrice je veux ajouter des tags aux articles afin de mieux les organiser.

- [x] - peut permettre de renseigner un ou plusieurs tags lors de la rédaction d'un article
- [x] - doit permettre l'autosuggestion de tags existants pour éviter les erreurs d'orthographe
- [x] - ne doit pas permettre d'enregistrer des tags de moins de deux caractères
- [x] - doit enregistrer le tag si celui-ci n'existe pas 
- [x] - doit permettre de voir les tags renseignés pour un article
- [x] - doit signaler les tags déjà sauvegardés
- [x] - doit comporter une animation à l'apparition
- [x] - doit permettre de supprimer un tag


cafebabel/articles/models.py
```diff
@@ -2,14 +2,33 @@
...
-from mongoengine import signals
+from mongoengine import PULL, signals
 
 from .. import db
 from ..core.helpers import slugify
 from ..core.mixins import UploadableImageMixin
 from ..users.models import User
 
 
+class Tag(db.Document, UploadableImageMixin):
```
vinyll: `UploadableImageMixin` requires the implementation of `get_images_url()` and `get_images_url()`

cafebabel/articles/models.py
```diff
@@ -2,14 +2,33 @@
...
+    name = db.StringField(required=True)
+    slug = db.StringField(required=True)
+    language = db.StringField(max_length=2, required=True)
+    summary = db.StringField()
+
+    def __str__(self):
+        return self.name
+
+    @classmethod
+    def update_slug(cls, sender, document, **kwargs):
```
vinyll: Would be nice to tests this

cafebabel/tests/test_articles.py
```diff
@@ -5,7 +5,7 @@
 from flask import url_for
 from flask.helpers import get_flashed_messages
 
-from ..articles.models import Article
+from ..articles.models import Article, Tag
```
vinyll: Add `Tag.drop_collection()` in `pytest_runtest_teardown`

vinyll (changes requested):

cafebabel/articles/models.py
davidbgk: Done.

cafebabel/articles/models.py
davidbgk: Done.

cafebabel/tests/test_articles.py
davidbgk: Done.

cafebabel/articles/models.py
```diff
@@ -17,12 +17,18 @@ class Tag(db.Document, UploadableImageMixin):
...
 
     @classmethod
     def update_slug(cls, sender, document, **kwargs):
         document.slug = slugify(document.name)
 
+    def get_images_url(self):
+        return current_app.config.get('TAGS_IMAGES_URL')
+
+    def get_images_path(self):
+        return current_app.config.get('TAGS_IMAGES_PATH')
```
vinyll: I guess it should therefore be [here too](https://github.com/cafebabel/cafebabel.com/blob/master/Makefile#L31).
Nevertheless, I know you were suggestion to ensure that the folders were created in another place, and I'm open to discussion.

davidbgk:
For now, tests are slow because all auth related fixtures are loaded for each test. We have to find a better way :)

cafebabel/__init__.py
```diff
@@ -66,22 +65,22 @@ def register_blueprints(app):
...
-        app.user_datastore.add_role_to_user(admin_user, admin_role)
-        click.echo('DB intialized.')
+        drop_collections()
+        auth_fixtures(app)
+        articles_fixtures(app)
+        tags_fixtures(app)
+        relations_fixtures(app)
+
+    @app.cli.command(short_help='Load articles and tags fixtures')
+    def load_data():
```
vinyll: `load_fixtures()`?

cafebabel/tests/test_articles.py
```diff
@@ -284,3 +284,35 @@ def test_article_to_translate_should_have_only_other_links(
...
     assert 'Translate in ' not in response
+
+
+def test_article_with_tag(app, tag, article):
+    Article.objects(id=article.id).update_one(push__tags=tag)
+    assert Article.objects(tags__in=[tag]).count() == 1
+    article2 = Article.objects(tags__in=[tag]).first()
+    assert article2.tags[0].summary == 'summary text'
+    assert tag.slug == 'a-tag'
+    assert str(tag) == 'A tag (en)'
```
vinyll: Feels like these 2 asserts testing the slug value and the stringification should be in a different test.

cafebabel/tests/test_articles.py
```diff
@@ -284,3 +284,35 @@ def test_article_to_translate_should_have_only_other_links(
...
+    Article.objects(id=article.id).update_one(pull__tags=tag)
+    assert Article.objects(tags__in=[tag]).count() == 0
+
+
+def test_article_with_tags(app, tag, article):
```
vinyll: Something more descsriptive like `test_article_deleting_tag_deletes_references`?

cafebabel/tests/test_articles.py
```diff
@@ -284,3 +284,35 @@ def test_article_to_translate_should_have_only_other_links(
...
+
+
+def test_article_detail_contains_tags(client, app, tag, published_article):
+    language = app.config['LANGUAGES'][0][0]
+    tag2 = Tag.objects.create(name='Tag two', language=language)
+    published_article.modify(tags=[tag, tag2])
+    response = client.get(
+        f'/article/{published_article.slug}-{published_article.id}/')
+    assert response.status_code == HTTPStatus.OK
+    assert 'A tag, Tag two.' in response
```
vinyll: This will very probably quickly break due to html adaptation.
Maybe you could just check the presence of each text individually?

cafebabel/tests/test_profile.py
```diff
@@ -31,36 +31,30 @@ def test_author_profile_has_list_of_published_articles_and_drafts(
...
-    login(client, editor.email, 'secret')
+    login(client, editor.email, 'password')
     response = client.get(f'/profile/{user.id}/edit/')
     assert response.status_code == HTTPStatus.OK
     assert not user.has_role('editor')
     assert 'name=editor' in response
-    response = client.post(f'/profile/{user.id}/edit/', data={
-        'editor': 'on',
-    })
+    client.post(f'/profile/{user.id}/edit/', data={'editor': 'on'})
```
vinyll: the possible values coming from the client side are `'1'` or `''`

cafebabel/tests/test_profile.py
```diff
@@ -31,36 +31,30 @@ def test_author_profile_has_list_of_published_articles_and_drafts(
...
-    login(client, user.email, 'secret')
+    login(client, user.email, 'password')
     response = client.get(f'/profile/{user.id}/edit/')
     assert response.status_code == HTTPStatus.OK
     assert not user.has_role('editor')
     assert 'name=editor' not in response
-    response = client.post(f'/profile/{user.id}/edit/', data={
-        'editor': 'on',
-    })
+    client.post(f'/profile/{user.id}/edit/', data={'editor': 'on'})
```
vinyll: same here

cafebabel/tests/test_profile.py
```diff
@@ -22,7 +22,7 @@ def test_user_profile_has_list_of_published_articles_no_draft(client, article):
 
 def test_author_profile_has_list_of_published_articles_and_drafts(
         client, user, article, published_article):
-    login(client, user.email, 'secret')
+    login(client, user.email, 'password')
     published_article.modify(author=user)
     response = client.get(f'/profile/', follow_redirects=True)
```
vinyll: `f` is useless here

cafebabel/tests/utils.py
```diff
@@ -1,8 +1,10 @@
 def login(client, email, password):
-    return client.post('/login', data=dict(
+    response = client.post('/login', data=dict(
         email=email,
         password=password
     ), follow_redirects=True)
+    assert b'<form action="/login"' not in response.data, response.data
```
vinyll: Why not `assert 'action="/login"' not in response` ?

cafebabel/tests/test_users.py
```diff
@@ -8,34 +8,26 @@
...
     assert response.status_code == HTTPStatus.FOUND
 
 
 def test_authenticated_user_can_access_login_required_page(client, user):
-    login(client, user.email, 'secret')
+    login(client, user.email, 'password')
     response = client.get('/profile/', follow_redirects=True)
     assert response.status_code == HTTPStatus.OK
-    assert b'<h1>testy@example.com\'s profile</h1>' in response.data
+    assert b'<h1>user@example.com\'s profile</h1>' in response.data
```
vinyll: `assert "<h1>user@example.com's profile</h1>" in response`?

cafebabel/tests/utils.py
davidbgk: We are out of the context where it works.

cafebabel/articles/drafts/views.py
```diff
@@ -24,8 +25,11 @@ def create():
 
     article = Article()
     authors = User.objects.all()
+    tags = {
```
vinyll: I'm wondering how many tags we may end up with

cafebabel/articles/drafts/views.py
```diff
@@ -37,8 +41,11 @@ def edit(draft_id):
         return redirect(article.detail_url)
 
     authors = User.objects.all()
+    tags = {
+        language[0]: Tag.objects(language=language[0])
+        for language in current_app.config['LANGUAGES']}
```
vinyll: As we have this twice, we may want to prefer writing a _staticmethod_ in Tag.

cafebabel/articles/models.py
```diff
@@ -115,8 +115,14 @@ def save_from_request(self, request):
                 del data['author']
             if data.get('editor'):
                 del data['editor']
+        self.tags = []
         for field, value in data.items():
-            setattr(self, field, value)
+            if field.startswith('tag-') and value:
+                tag = Tag.objects.get(name=value)
```
vinyll: Shouldn't you filter the language here?

cafebabel/templates/articles/macros.html
```diff
@@ -15,3 +15,19 @@
...
+{%- macro tags_fields(tags, article=None) -%}
+  <div>
+    <label for=tag-1>Tags</label>
+    <input list=tags name=tag-1 {% if article and article.tags and article.tags[0] %}value={{ article.tags[0].name }}{% endif %}>
+    <input list=tags name=tag-2 {% if article and article.tags and article.tags[1] %}value={{ article.tags[1].name }}{% endif %}>
+    <input list=tags name=tag-3 {% if article and article.tags and article.tags[2] %}value={{ article.tags[2].name }}{% endif %}>
+    <input list=tags name=tag-4 {% if article and article.tags and article.tags[3] %}value={{ article.tags[3].name }}{% endif %}>
+    <input list=tags name=tag-5 {% if article and article.tags and article.tags[4] %}value={{ article.tags[4].name }}{% endif %}>
+    <datalist id=tags>
+      {% for tag in tags['en'] %}
```
vinyll: Why filtering for English tags instead of the article's language?

cafebabel/templates/articles/macros.html
```diff
@@ -17,13 +17,16 @@
...
-    <input list=tags name=tag-1 {% if article and article.tags and article.tags[0] %}value={{ article.tags[0].name }}{% endif %}>
-    <input list=tags name=tag-2 {% if article and article.tags and article.tags[1] %}value={{ article.tags[1].name }}{% endif %}>
-    <input list=tags name=tag-3 {% if article and article.tags and article.tags[2] %}value={{ article.tags[2].name }}{% endif %}>
-    <input list=tags name=tag-4 {% if article and article.tags and article.tags[3] %}value={{ article.tags[3].name }}{% endif %}>
-    <input list=tags name=tag-5 {% if article and article.tags and article.tags[4] %}value={{ article.tags[4].name }}{% endif %}>
+    <div class=tag-container><input name=tag-new><button class=add></button></div>
+    {% if article and article.tags %}
+    {% set count = 1 %}
+      {% for tag in tags['en'] %}
+        <button class=remove>+</button><input list=tags name=tag-{{ count }} value={{ article.tags.name }}>
```
vinyll: you may need the quotes for the value here at it may contain space or other special character

vinyll: From a code point of view – without the interface – it looks surprising to have a "+" with a class _remove_

cafebabel/templates/articles/macros.html
```diff
@@ -17,13 +17,16 @@
...
-  <div>
+  <div class=tags>
     <label for=tag-1>Tags</label>
-    <input list=tags name=tag-1 {% if article and article.tags and article.tags[0] %}value={{ article.tags[0].name }}{% endif %}>
-    <input list=tags name=tag-2 {% if article and article.tags and article.tags[1] %}value={{ article.tags[1].name }}{% endif %}>
-    <input list=tags name=tag-3 {% if article and article.tags and article.tags[2] %}value={{ article.tags[2].name }}{% endif %}>
-    <input list=tags name=tag-4 {% if article and article.tags and article.tags[3] %}value={{ article.tags[3].name }}{% endif %}>
-    <input list=tags name=tag-5 {% if article and article.tags and article.tags[4] %}value={{ article.tags[4].name }}{% endif %}>
+    <div class=tag-container><input name=tag-new><button class=add></button></div>
+    {% if article and article.tags %}
```
vinyll: `{% if … article.tags %}`: does it still work when creating an article or editing an article with no tag, or will these fields not appear at all?

cafebabel/static/js/edit-article.js
```diff
@@ -0,0 +1,6 @@
+const tagButtonAdd = document.querySelector('.tags button.add')
+const tagButtonRemove = document.querySelector('.tags button.remove')
+tagButtonAdd.addEventListener('click', (event) => {
```
vinyll: I see this is stopping the native behavior, but it seems like it's doing nothing instead.
Ane what is the default behavior of a button anyways?

cafebabel/templates/articles/macros.html
```diff
@@ -15,3 +15,22 @@
     Reading time: {{ article.body|reading_time }} min
   </p>
 {%- endmacro %}
+
+{%- macro tags_fields(tags, article=None) -%}
+  <div class=tags>
+    <label for=tag-1>Tags</label>
+    <div class=tag-container><input name=tag-new><button class=add>+</button></div>
+    {% if article and article.tags %}
+    {% set count = 1 %}
```
davidbgk: I think you are looking for [`loop.index`](http://jinja.pocoo.org/docs/2.10/templates/#for) 😉

cafebabel/static/js/edit-article.js
```diff
@@ -1,6 +1,28 @@
 const tagButtonAdd = document.querySelector('.tags button.add')
 const tagButtonRemove = document.querySelector('.tags button.remove')
+const tagsList = document.querySelector('.tags .tagsList')
+
+const Tag = {
+  count: () => tagsList.querySelectorAll('li').length,
```
vinyll: I think a real method would seem more appropriate
```
count () {
  return tagsList.querySelectorAll('li').length
}
```

cafebabel/static/js/edit-article.js
```diff
@@ -1,6 +1,28 @@
...
+  createTag(tagValue) {
```
vinyll: missing space before `(`

cafebabel/static/js/edit-article.js
```diff
@@ -1,6 +1,28 @@
...
+    li.append(tagValue)
+    input.type = 'hidden'
+    input.name = `tag-${Tag.count()}`
+    input.setAttribute('list', 'tags')
+    input.value = tagValue
+    li.append(input)
+
+    return li
+  },
+  tagValues: () => Array.from(tagsList.querySelectorAll('li')).map(li => li.innerText)
```
vinyll: same here

cafebabel/templates/articles/macros.html
```diff
@@ -20,13 +20,16 @@
   <div class=tags>
     <label for=tag-1>Tags</label>
     <div class=tag-container><input name=tag-new><button class=add>+</button></div>
-    {% if article and article.tags %}
-    {% set count = 1 %}
+    <ul class=tagsList>
+      {% set count = 1 %}
       {% for tag in tags['en'] %}
```
vinyll: Still wondering why we take English only here

cafebabel/templates/articles/macros.html
```diff
@@ -20,13 +20,16 @@
   <div class=tags>
     <label for=tag-1>Tags</label>
     <div class=tag-container><input name=tag-new><button class=add>+</button></div>
-    {% if article and article.tags %}
-    {% set count = 1 %}
+    <ul class=tagsList>
```
vinyll: We haven't used camelCased class names so far but hyphen separated.
Let's try to be consistent

vinyll (changes requested):

cafebabel/static/js/edit-article.js
```diff
@@ -2,5 +2,10 @@ const tagButtonAdd = document.querySelector('.tags button.add')
 const tagButtonRemove = document.querySelector('.tags button.remove')
 tagButtonAdd.addEventListener('click', (event) => {
   event.preventDefault()
+  const tag = event.target.previousElementSibling.value
+  if (!tag) {
+    return
+  }
+  console.log('Tag', tag)
```
vinyll: You may remove this

cafebabel/static/js/edit-article.js
```diff
@@ -2,5 +2,10 @@ const tagButtonAdd = document.querySelector('.tags button.add')
 const tagButtonRemove = document.querySelector('.tags button.remove')
 tagButtonAdd.addEventListener('click', (event) => {
   event.preventDefault()
+  const tag = event.target.previousElementSibling.value
+  if (!tag) {
```
vinyll: A online liner is appropriate here: `if (!tag) return`

vinyll (changes requested):

cafebabel/templates/articles/macros.html
```diff
@@ -20,13 +20,14 @@
...
+      <li>Machin<input type=hidden list=tags name=tag-0 value=machin></li>
```
vinyll: `value=machin` 🙃

vinyll (changes requested):

cafebabel/static/js/edit-article.js
```diff
@@ -13,15 +13,16 @@ const Tag = {
...
-  }
+  },
+  tagValues: () => Array.from(tagsList.querySelectorAll('li')).map(li => li.innerText)
 }
 
 tagButtonAdd.addEventListener('click', (event) => {
   event.preventDefault()
   const tagValue = event.target.previousElementSibling.value
-  if (!tagValue) return
+  if (!tagValue || Tag.tagValues().includes(tagValue)) return
```
vinyll: Just a suggestion: because you have created a Tag object, it may be interesting to create a method to check `Tag.contains(tagValue)`.

cafebabel/static/js/edit-article.js
```diff
@@ -16,13 +16,21 @@ const Tag = {
 
     return li
   },
-  tagValues: () => Array.from(tagsList.querySelectorAll('li')).map(li => li.innerText)
+  tagValues: () => Array.from(tagsList.querySelectorAll('li')).map(li => li.innerText),
+  createTagsList: (container, tagValues) => {
```
vinyll: `createTagsList (container, tagValues) {`

cafebabel/static/js/edit-article.js
```diff
@@ -16,13 +16,21 @@ const Tag = {
...
 }
 
 tagButtonAdd.addEventListener('click', (event) => {
   event.preventDefault()
-  const tagValue = event.target.previousElementSibling.value
-  if (!tagValue || Tag.tagValues().includes(tagValue)) return
-  tagsList.appendChild(Tag.createTag(tagValue))
+  const tag = event.target.previousElementSibling.value
+  if (!tag || Tag.tagValues().includes(tag)) return
+  const tags = Tag.tagValues().concat(tag)
```
vinyll: Same for creating a method here. It could be `Tag.add(tag)`

cafebabel/static/js/edit-article.js
```diff
@@ -1,36 +1,48 @@
-const tagButtonAdd = document.querySelector('.tags button.add')
-const tagButtonRemove = document.querySelector('.tags button.remove')
-const tagsList = document.querySelector('.tags .tagsList')
+class Tags {
+  constructor(domContext) {
```
vinyll: the convention that has been taken in this project is to have a space after the name when declaring a function name. Let's keep this uniform.

cafebabel/static/js/edit-article.js
```diff
@@ -1,36 +1,48 @@
...
+    this.list = domContext.querySelector('ul')
+    this.values = Array.from(this.list.querySelectorAll('li')).map(li => li.innerText)
+    this.submission = domContext.querySelector('input[name=tag-new]').value
+  }
+  checkSubmission() {
```
vinyll: same

cafebabel/static/js/edit-article.js
```diff
@@ -1,36 +1,48 @@
...
+    return this.submission && ! this._isTag(this.submission)
+  }
+  display() {
```
vinyll: and here

cafebabel/static/js/edit-article.js
```diff
@@ -1,36 +1,48 @@
...
+    const tagsValue = this.values.concat(this.submission)
+    const container = this.list.cloneNode(false)
+    this.list.replaceWith(this._createTagsList(container, tagsValue))
+  }
 
-const Tag = {
-  count: () => tagsList.querySelectorAll('li').length,
-  createTag(tagValue) {
+  _isTag(tag) {
```
vinyll: again

cafebabel/static/js/edit-article.js
```diff
@@ -1,36 +1,48 @@
...
+    return this.values.includes(tag)
+  }
+  _createTagsList(container, tagValues) {
```
vinyll: here too

cafebabel/static/js/edit-article.js
```diff
@@ -1,36 +1,48 @@
...
+    tagValues.forEach((tagValue, index) => container.appendChild(this._createTag(tagValue, index)))
+
+    return container
+  }
+  _createTag(tagValue, index) {
```
vinyll: and here

cafebabel/static/js/edit-article.js
vinyll: I'd suggest calling it `render` as it seems to become more of a convention.
But this is subjective, up to you ;)

vinyll (changes requested):

cafebabel/static/js/edit-article.js
```diff
@@ -2,17 +2,16 @@ class Tags {
   constructor(domContext) {
     this.list = domContext.querySelector('ul')
     this.values = Array.from(this.list.querySelectorAll('li')).map(li => li.innerText)
-    this.submission = domContext.querySelector('input[name=tag-new]').value
   }
-  checkSubmission() {
-    return this.submission && ! this._isTag(this.submission)
+  checkSubmission(query) {
```
vinyll: need a space after function name

cafebabel/static/js/edit-article.js
```diff
@@ -31,18 +30,35 @@ class Tags {
...
+  const submission = event.target.previousSibling.value
   const tags = new Tags(document.querySelector('.tags'))
-  if (!tags.checkSubmission()) return
-  tags.display()
+  if (tags.checkSubmission(submission)) tags.add(submission)
+  displayTags(tags)
+  event.target.previousSibling.value = ''
 })
 
+function addTagsRemoveListener() {
```
vinyll: add a space after function name

cafebabel/static/js/edit-article.js
```diff
@@ -31,18 +30,35 @@ class Tags {
...
+  const tagsButtonRemove = document.querySelectorAll('.tags ul')
+  tagsButtonRemove.forEach(tagButtonRemove => tagButtonRemove.addEventListener('click', (event) => {
+    const tags = new Tags(document.querySelector('.tags'))
+    tags.remove(event.target.innerText)
+    displayTags(tags)
+  }))
+}
+
+function displayTags(tags) {
```
vinyll: same here

vinyll (changes requested):

cafebabel/articles/models.py
davidbgk: Done.

cafebabel/templates/articles/macros.html
```diff
@@ -15,3 +15,12 @@
     Reading time: {{ article.body|reading_time }} min
   </p>
 {%- endmacro %}
+
+{%- macro tags_fields(tags, article=None) -%}
+  <div class=tags>
+    <label for=tag-1>Tags</label>
+    <ul class=tags-list></ul>
```
davidbgk: @cafebabel-webmaster I wasn't sure of the purpose of that list, maybe useless now we have the datalist for suggestion?

cafebabel/templates/articles/macros.html
```diff
@@ -15,3 +15,21 @@
     Reading time: {{ article.body|reading_time }} min
   </p>
 {%- endmacro %}
+
+{%- macro tags_fields(article=None) -%}
+  <div class=tags>
+    <label for=tag-1>Tags</label>
+    <ul class=tags-list>
```
davidbgk, line 22: Here we use `.tags-list` for style and `.tags ul` on the JS side. We have to be consistent @pointbar @cafebabel-webmaster

cafebabel/static/js/edit-article.js
pointbar: @vinyll - I think I'm missing something here.
Do you ask to add a space between function name and args ?
Like this:
 constructor (domContext) - instead of: constructor(domContext)

davidbgk: It is now handled by prettierjs

cafebabel/static/js/edit-article.js
```diff
@@ -89,6 +89,6 @@ window.addEventListener('load', addTagsRemoveListener)
 function displayTags(tags) {
   const container = tags.list.cloneNode(false)
   tags.list.replaceWith(tags._createTagsList(container, tags.values))
-  document.querySelector('.tags ul li:last-child').classList.add('fadeIn')
+  document.querySelector('.tags .tags-list li:last-child').classList.add('fadeIn')
```
vinyll: Not sure why adding yet another class within such a specific `.tags` class

pointbar: @vinyll - @johan creates another -ul- for completion, is your comment stays relevant with this new context?

johan: Johan != @johan on github, FYI, @pointbar. :) / disables notifications.

cafebabel/templates/articles/macros.html
```diff
@@ -15,3 +15,21 @@
...
+        {% for tag in article.tags -%}
+          <li>
+            {{ tag.name }}
+            <input name=tag-{{ loop.index }} list=tags value="{{ tag.name }}" type=hidden>
+          </li>
+        {%- endfor %}
+      {% endif %}
+    </ul>
+    <div class=tag-container><input name=tag-new autocomplete=off list=tags-suggestions><button class=add>+</button></div>
+    <ul id=tags-suggestions class=inactive></ul>
```
davidbgk, line 34: @cafebabel-webmaster changing that `datalist` element to a `ul` probably broke the native autocomplete from the browser, not sure that is something we want? https://developer.mozilla.org/en-US/docs/Web/HTML/Element/datalist

johangiraud: I couldn't style the datalist elements. That's why I switched for `ul` and `li`. we can talk about it later

cafebabel/static/js/edit-article.js
```diff
@@ -0,0 +1,146 @@
+class Tags {
+  constructor() {
+    const domContext = document.querySelector('.tags')
+    this.list = domContext.querySelector('.tags-list')
+    this.values = Array.from(this.list.querySelectorAll('input')).map(
+      input => input.value
+    )
+  }
+  static addTagsRemoveListener() {
+    const tagsButtonRemove = document.querySelectorAll('.tags .tags-list li a')
```
davidbgk: Reuse `this.list` here?

cafebabel/static/js/edit-article.js
```diff
@@ -0,0 +1,146 @@
...
+  }
+  _createTag(tagValue, index) {
+    const li = document.createElement('li')
+    const input = document.createElement('input')
+    const a = document.createElement('a')
+    input.type = 'hidden'
+    input.name = `tag-${index + 1}`
+    input.setAttribute('list', 'tags')
+    input.value = tagValue
+    li.append(tagValue)
```
davidbgk: I think you can do `li.append(tagValue, input, a)`

cafebabel/static/js/edit-article.js
```diff
@@ -0,0 +1,146 @@
...
+    li.append(a)
+
+    return li
+  }
+}
+
+class Effect {
+  static fadeIn(element) {
+    element.style.opacity = 0
+    ;(function fade() {
```
davidbgk: Out of curiosity, why IIFEs here and below?

cafebabel/static/js/edit-article.js
```diff
@@ -27,68 +39,108 @@ class Tags {
...
     const li = document.createElement('li')
     const input = document.createElement('input')
+    const a = document.createElement('a')
     input.type = 'hidden'
     input.name = `tag-${index + 1}`
     input.setAttribute('list', 'tags')
     input.value = tagValue
     li.append(tagValue)
     li.append(input)
+    li.append(a)
```
vinyll: `a` seems like blank with no attribute. Is it normal?

We could also consider a JSX-style for replacing the ~10 lines above – which might be more readable:
```js
return `
  <li>
    <input type=hidden list=tags name=tag-${index + 1} value="${tagValue}">
    <a></a>
  </li>
`
```

cafebabel/static/js/edit-article.js
```diff
@@ -27,68 +39,108 @@ class Tags {
...
-    document.querySelector('#tags-suggestions').innerHTML = ''
-    const submission = event.target.value
-    const languages = document.querySelector('#language')
-    const language = languages.options[languages.selectedIndex].value
-    if (submission.length < 3) return
-    request(`/article/tag/suggest/?language=${language}&terms=${submission}`)
-      .then(response => response.forEach(handleSuggestion))
-      .catch(console.error.bind(console))
-  })
+class Listener {
```
vinyll: This seems very specific to tags while the class sounds very generic.
Why not going for a name like `TagEventListener`?

cafebabel/static/js/edit-article.js
```diff
@@ -27,68 +39,108 @@ class Tags {
...
+    })
+  }
 }
+
+Listener.clickAdd()
+Listener.keyup()
+
+window.addEventListener('load', () => {
+  const tags = new Tags(document.querySelector('.tags'))
+  Tags.addTagsRemoveListener()
```
vinyll: This `addTagsRemoveListener` method should probably belong to the currently called `Listener` class?

cafebabel/static/js/edit-article.js
```diff
@@ -142,5 +138,5 @@ Listener.keyup()
 
 window.addEventListener('load', () => {
   const tags = new Tags(document.querySelector('.tags'))
```
davidbgk: If I understand correctly the constructor of `Tags` doesn't take any parameter now?

cafebabel/templates/articles/drafts/create.html
```diff
@@ -1,6 +1,8 @@
 {% extends "base.html" %}
 {% import "articles/macros.html" as macros without context %}
 
+{% block extra_js %}<script src={{ url_for('static', filename='js/edit-article.js') }}></script>{% endblock %}
```
davidbgk: Maybe we should better name it `articles-tags.js`, it's kind of weird to have an edit file in the create template 😉

cafebabel/static/js/edit-article.js
pointbar: Not fine with the old version

cafebabel/static/js/edit-article.js
```diff
@@ -0,0 +1,147 @@
...
+    tagValues.forEach((tagValue, index) =>
+      container.appendChild(this._createTag(tagValue, index))
+    )
+
+    return container
+  }
+  _createTag(tagValue, index) {
+    const li = document.createElement('li')
+    const input = document.createElement('input')
+    const a = document.createElement('a')
```
davidbgk: I second @vinyll here, semantically it's closer to a `button` element. Thoughts?

cafebabel/static/js/edit-article.js
```diff
@@ -0,0 +1,147 @@
...
+class Listener {
+  static clickAdd() {
+    const tagButtonAdd = document.querySelector('.tags button.add')
+    tagButtonAdd.addEventListener('click', event => {
+      event.preventDefault()
+      const submission = event.target.previousSibling.value
+      const tags = new Tags()
+      if (tags.checkSubmission(submission)) tags.add(submission)
+      tags.display()
+      Effect.fadeIn(document.querySelector('.tags .tags-list li:last-child'))
```
davidbgk: Reuse `tags.list`?

cafebabel/static/js/edit-article.js
```diff
@@ -0,0 +1,147 @@
...
+      document.querySelector('.tags input[name=tag-new]').value = ''
+    })
+  }
+  static clickRemove(tagButtonRemove) {
+    tagButtonRemove.addEventListener('click', event => {
+      const tags = new Tags()
+      const tagElement = event.target.parentNode
+      tags.remove(tagElement.innerText)
+      Effect.fadeOut(tagElement).then(response => tags.display())
```
davidbgk: I think you don't have a `response` here given that you don't `resolve()` any value, I would just do `.then(tags.display)`

cafebabel/static/js/edit-article.js
pointbar: We don't find any solution to attach event on the pseudo-element ::after.
After tries we add an `a` to attach a -remove link-

cafebabel/static/js/articles-tags.js
```diff
@@ -0,0 +1,150 @@
...
+        this._createTag(tagValue, index)
+      )
+    )
+
+    return container
+  }
+  _createTag(tagValue, index) {
+    return `
+      <li>
+        ${tagValue}
```
davidbgk: Bad news, with that implementation [extra space bug](https://github.com/cafebabel/cafebabel.com/issues/151) stroked again, this time on the JS side 🙃

johangiraud: We wrote it as one only line, and it fixes it.

cafebabel/static/js/articles-tags.js
```diff
@@ -2,13 +2,15 @@ class Tags {
   constructor() {
     const context = document.querySelector('.tags')
     this.list = context.querySelector('.tags-list')
-    this.buttonAdd = context.querySelector('input[name=tag-new]')
+    this.fieldAdd = context.querySelector('input[name=tag-new]')
+    this.buttonAdd = context.querySelector('button.add')
+    this.removeButtons = context.querySelectorAll('.tags-list li button')
```
vinyll: `querySelectorAll` returns a _NodeList_.
In some browsers, `forEach` is not implemented on a NodeList.
It may be safer to convert it into an _array_: `this.removeButtons = Array.from(context.querySelectorAll('.tags-list li button'))`

cafebabel/static/js/articles-tags.js
```diff
@@ -52,7 +80,7 @@ class TagEffect {
         requestAnimationFrame(fade)
       }
     }
-
+    console.log('TOC FADE')
```
davidbgk: To be removed

cafebabel/static/js/articles-tags.js
```diff
@@ -0,0 +1,166 @@
...
+    inputNewTag.addEventListener('keypress', event => {
+      /* Return if arrow up, arrow down or enter are pressed */
+      if (event.keyCode == 13) {
+        return
+      }
+      if (event.keyCode == 40 || event.keyCode == 38) return
+      const submission = event.target.value
+      if (submission.length < 3) return
+      const tags = new Tags()
+      request(`/article/tag/suggest/?language=en&terms=${submission}`)
```
davidbgk: Why removing the dynamic language choice?!

pointbar: @davidbgk - sorry, I misunderstood something about languages

cafebabel/static/js/articles-tags.js
```diff
@@ -0,0 +1,180 @@
...
+    return this.context.querySelector('#tags-suggestions')
+  }
+  get buttonAdd() {
+    return this.context.querySelector('button.add')
+  }
+  get fieldAdd() {
+    return this.context.querySelector('input[name=tag-new]')
+  }
+  get removeButtons() {
+    return this.context.querySelectorAll('.tags-list li button')
```
davidbgk: Reuse `this.list()`?

cafebabel/static/js/articles-tags.js
davidbgk: Where/when? The current implementation still adds it.

cafebabel/static/js/articles-tags.js
```diff
@@ -0,0 +1,180 @@
...
+        const tagElement = event.target.parentNode
+        tags.removeTag(tagElement)
+      })
+    )
+  }
+  static keyup() {
+    const inputNewTag = document.querySelector('.tags input[name=tag-new]')
+    inputNewTag.addEventListener('keypress', event => {
+      /* Return if arrow up, arrow down or enter are pressed */
+      if (event.keyCode == 13) {
```
davidbgk: Why the recent split with arrows detection?

cafebabel/static/js/articles-tags.js
```diff
@@ -0,0 +1,180 @@
...
+        return
+      }
+      if (event.keyCode == 40 || event.keyCode == 38) return
+      const submission = event.target.value
+      if (submission.length < 3) return
+      const tags = new Tags()
+      request(
+        `/article/tag/suggest/?language=${tags.language}&terms=${submission}`
+      )
+        .then(tagsApi => tags.handleSuggestion(tagsApi))
```
davidbgk: Can be simplified as `.then(tags.handleSuggestion)`

cafebabel/static/js/articles-tags.js
pointbar: Yes, but`this.list`;)

cafebabel/static/js/articles-tags.js
pointbar: I try to resolve the problem with -keycode: 13- 
Can I remove this case, this `addEventListener`never captures -return key-

cafebabel/static/js/articles-tags.js
pointbar: @davidbgk - I feel like I'm losing `this` and the argument?

davidbgk: Ah! Maybe not that simple 👌

cafebabel/static/js/articles-tags.js
```diff
@@ -69,13 +69,11 @@ class Tags {
     return container
   }
   _createTag(tagValue, index) {
-    return `
-      <li>
+    return `<li>
         ${tagValue}
         <input name="tag-${++index}" list="tags" value=${tagValue} type="hidden">
```
vinyll: I suppose you need quotes nowhere but on the `value` attribute as it's the one that may have spaces or special chars 😅

vinyll (changes requested):

cafebabel/static/js/articles-tags.js
```diff
@@ -40,14 +40,18 @@ class Tags {
...
-    const ul = this.suggestions.cloneNode(false)
-    tagsApi.forEach(tag => {
-      const li = `<li>${tag.name}</li>`
-      ul.insertAdjacentHTML('afterbegin', li)
-    })
-    this._renderSuggestion(ul)
+  handleSuggestion(submission) {
+    _request(submission)
+      .then(tagsApi => this._createSuggestionList(tagsApi))
+      .catch(console.error.bind(console))
```
vinyll: Should we alert the user if an error occurs?

cafebabel/static/js/articles-tags.js
```diff
@@ -40,14 +40,18 @@ class Tags {
...
+  }
+  _request(submission) {
+    return request(
+      `/article/tag/suggest/?language=${this.language}&terms=${submission}`
+    )
+  }
+  _isTagSaved(tag) {
+    return
```
vinyll: 🙃

cafebabel/static/js/articles-tags.js
```diff
@@ -174,12 +169,12 @@ class TagEventListener {
       if (event.keyCode == 40 || event.keyCode == 38) return
```
vinyll: Could you code-comment what these keyCodes are?
Eventually also _why_ to prevent key up and down as it's not obvious from here.

cafebabel/static/js/articles-tags.js
pointbar: I am writing on

cafebabel/static/js/articles-tags.js
```diff
@@ -48,8 +48,10 @@ class Tags {
       `/article/tag/suggest/?language=${this.language}&terms=${submission}`
     )
   }
-  _isTagSaved(tag) {
-    return
+  _isTagSaved(submission) {
+    return this._request(submission).then(
+      tagsApi => !!tagsApi.filter(tagApi => tagApi.name === submission).length
```
vinyll: Unless I missunderstood, it would be more straightforward to write:
```js
tagsApi => tagsApi.some(tag => tag.name === submission)
```

cafebabel/static/js/articles-tags.js
vinyll: It seems obvious to you, but _why_ do we need to block them?

cafebabel/templates/articles/detail.html
```diff
@@ -34,11 +34,10 @@
     <h1>{{ article.title }}</h1>
 
     {{ macros.article_meta(article) }}
-
-    <p>
-      Tags:
+    <p class=tags-list>
       {% for tag in article.tags -%}
-        {{ tag.name }}{% if loop.last %}.{% else %}, {% endif %}
+        {#<a href={{ url_for('tags.detail', slug=tag.slug) }}>{{ tag.name }}</a>#} <!-- next PR -->
```
vinyll: `<!-- next PR -->`? 🙃

cafebabel/templates/articles/detail.html
```diff
@@ -34,13 +34,14 @@
...
 
     {{ macros.article_meta(article) }}
-
-    <p>
-      Tags:
-      {% for tag in article.tags -%}
-        {{ tag.name }}{% if loop.last %}.{% else %}, {% endif %}
-      {%- endfor %}
-    </p>
+    {% if article.tags|length != 0 %}
```
davidbgk: `if article.tags` is enough.

cafebabel/templates/articles/detail.html
```diff
@@ -34,13 +34,14 @@
...
+      <p class=tags-list>
+        {% for tag in article.tags -%}
+          {#<a href={{ url_for('tags.detail', slug=tag.slug) }}>{{ tag.name }}</a>#} <!-- next PR -->
```
davidbgk: It has nothing to do here.

cafebabel/templates/articles/detail.html
```diff
@@ -34,13 +34,14 @@
...
+          {{ tag.name }}
```
davidbgk: Please get back to the previous version: `{{ tag.name }}{% if loop.last %}.{% else %}, {% endif %}` or change tests (or explain why you did that in the commit message)

cafebabel/templates/articles/detail.html
johangiraud: oui ça sera la ligne à décommenter dans la prochaine PR

cafebabel/templates/articles/detail.html
johangiraud: ok

cafebabel/templates/articles/detail.html
johangiraud: I want to have that in the end: capture d ecran 2018-01-18 a 16 30 02 so I don't need commas or dot

davidbgk: I know, that's just not the right branch/PR.

vinyll (dismissed):

cafebabel/static/js/articles-tags.js
```diff
@@ -171,8 +170,10 @@ class TagEventListener {
   static keyup() {
     const inputNewTag = document.querySelector('.tags input[name=tag-new]')
     inputNewTag.addEventListener('keyup', event => {
+      event.preventDefault()
       /* Return if arrow up, arrow down are pressed */
-      if (event.keyCode == 40 || event.keyCode == 38) return
+      if (event.keyCode == 13 || event.keyCode == 40 || event.keyCode == 38)
```
vinyll: Arrow up, down… and enter?
It would be nice to tell _why_ this hack needs to be processed in very short.

vinyll: Comment should be updated accordingly

johangiraud (dismissed):

johangiraud (approved):

PR #147, opened 2018-01-11
82 - Ajouter les méta données dans les différentes pages

johangiraud:
Ajouter les méta données dans les différentes pages pour les réseaux sociaux Open graph & Twitter cards


johangiraud:
question: og:image needs a absolute url. Thing is I have only a relative one. How could I get the full url of the cover image ? or complete the short one I have @davidbgk

johangiraud:
another one: how can I get the current interface language in base.html? @davidbgk

davidbgk:
@cafebabel-webmaster 

> question: og:image needs a absolute url. Thing is I have only a relative one. How could I get the full url of the cover image ? or complete the short one I have

I will amend the PR to deal with that.

> another one: how can I get the current interface language in base.html?

Languages support is not implemented yet.

cafebabel/templates/base.html
```diff
@@ -3,11 +3,28 @@
...
+<meta name="twitter:title" content="{{ self.meta_title() }}">
+<meta name="twitter:description" content="{{ self.meta_description() }}">
+<meta name="twitter:image" content="{{ self.meta_image_url() }}">
+<meta name="twitter:url" content="{{ request.url }}">
 
 <link rel=stylesheet href=/static/css/styles.css>
 <link rel=stylesheet href=/static/css/edit.css>
 <script src=/static/js/utils.js></script>
 
+{% block js %}{% endblock %}
```
davidbgk: Not in use?

cafebabel/templates/base.html
```diff
@@ -3,11 +3,28 @@
...
+<meta property="og:description" content="{{ self.meta_description() }}">
+<meta property="og:site_name" content="Cafébabel">
+<meta property="og:locale" content="{% block meta_language %}{% endblock %}">
+<meta property="article:author" content="{% block meta_author %}{% endblock %}">
+
+<meta name="twitter:card" content="summary">
+<meta name="twitter:title" content="{{ self.meta_title() }}">
+<meta name="twitter:description" content="{{ self.meta_description() }}">
+<meta name="twitter:image" content="{{ self.meta_image_url() }}">
+<meta name="twitter:url" content="{{ request.url }}">
```
davidbgk: You *might* want to [factorize all that metacrap](https://adactio.com/journal/9881) but it costs some readability, I let you arbitrate :)

davidbgk:
@cafebabel-webmaster : can you update the title/description of the PR?

cafebabel/templates/base.html
johangiraud: set in PR 14-profile-image. I remove it from this branch to avoid conflicts

johangiraud:
@davidbgk I updates the description/title

davidbgk (dismissed):
There is only the refactoring of meta left to you @cafebabel-webmaster (and maybe removing the `to discuss` comment?)

cafebabel/core/mixins.py
```diff
@@ -17,7 +19,15 @@ def get_images_path(self):
     @property
     def image_url(self):
         if self.has_image:
-            return f'{self.get_images_url()}/{self.get_id()}'
+            return url_for('static',
+                           filename=f'{self.get_images_url()}/{self.get_id()}')
+
+    @property
+    def image_full_url(self):
```
vinyll: I would rename it to `image_absolute_url` if we had to do this that way for some reason.
However I'd suggest to use a `absolute()` template function instead.

cafebabel/templates/base.html
```diff
@@ -3,6 +3,21 @@
...
 <title>{% block meta_title %}pan-European news, features and opinions{% endblock %} – Cafebabel</title>
 <meta name=description content="{% block meta_description %}{% endblock %}">
+<!--Social-->
+<meta property="og:url" content="{{ request.url }}">
+<meta property="og:title" content="{{ self.meta_title() }}">
+<meta property="og:image" content="{% block meta_image_url %}{% endblock %}"> <!-- to discuss -->
+<meta property="og:type" content="{% block meta_type %}website{% endblock %}">
+<meta property="og:description" content="{{ self.meta_description() }}">
+<meta property="og:site_name" content="Cafébabel">
+<meta property="og:locale" content="{% block meta_language %}{% endblock %}">
```
vinyll: Not sure if this will be manage on a per-block basis or per-website version.
We should probably detect the language based on the domain (fr: cafebabel.fr, en: cafebabel.co.uk, …).

cafebabel/templates/base.html
```diff
@@ -3,6 +3,21 @@
 <meta name=viewport content="width=device-width, minimum-scale=1, initial-scale=1, shrink-to-fit=no">
 <title>{% block meta_title %}pan-European news, features and opinions{% endblock %} – Cafebabel</title>
 <meta name=description content="{% block meta_description %}{% endblock %}">
+<!--Social-->
+<meta property="og:url" content="{{ request.url }}">
+<meta property="og:title" content="{{ self.meta_title() }}">
+<meta property="og:image" content="{% block meta_image_url %}{% endblock %}"> <!-- to discuss -->
```
vinyll: What is the `<!-- to discuss -->` for?
If it's a comment for development, please use `{# to discuss #}` instead so it does not go into the response HTML

cafebabel/tests/test_articles.py
```diff
@@ -47,6 +47,8 @@ def test_published_article_should_display_content(client, published_article,
             f'localhost%2Farticle%2F{published_article.slug}-'
             f'{published_article.id}%2F"' in response)
     assert '1 min' in response
+    assert (f'<meta property="og:url" content="http://localhost/article/'
```
vinyll: no need of `f` here

cafebabel/tests/test_articles.py
```diff
@@ -134,6 +136,8 @@ def test_update_article_with_image_should_return_200(app, client, user, editor,
     assert published_article.has_image
     assert (Path(app.config.get('ARTICLES_IMAGES_PATH') /
```
vinyll: useless perenthesis

cafebabel/tests/test_articles.py
```diff
@@ -134,6 +136,8 @@ def test_update_article_with_image_should_return_200(app, client, user, editor,
...
                  str(published_article.id)).exists())
+    assert (f'<meta property="og:image" content="http://localhost/static/'
```
vinyll: no need of `f`

cafebabel/__init__.py
```diff
@@ -103,9 +103,9 @@ def register_context_processors(app):
 
     @app.context_processor
     def add_template_helpers():
+        from .core.helpers import current_language
         return dict(
             get_languages=lambda: app.config.get('LANGUAGES', tuple()),
             get_categories=lambda: app.config.get('CATEGORIES', []),
-            article_image_url=(
-                lambda a: f'{app.config.get("ARTICLES_IMAGES_URL")}/{a.id}')
+            current_language=current_language(),
```
davidbgk: Are you sure that it is not cached across requests (hence the use of lambdas)?

config.py
```diff
@@ -59,15 +59,35 @@ class BaseConfig:
         'es': '@'.join(['redaccion', 'cafebabel.com']),
         'it': '@'.join(['redazione', 'cafebabel.com']),
     }
-    ARTICLES_IMAGES_URL = '/static/uploads/articles'
+
+    DOMAIN_LANGUAGES = {
```
davidbgk, line 63: Does it override `SERVER_NAME`? (in use for instance with `url_for` + `_external`) How do we test that?

davidbgk: Which makes me think that if we want to be able to link translations across domains we will have to deal with a custom `url_for`. Oh and we need to discuss how we store/retrieve uploads across languages too! ;)

cafebabel/templates/base.html
johangiraud: Ok thanks !

cafebabel/templates/base.html
```diff
@@ -3,6 +3,21 @@
 <meta name=viewport content="width=device-width, minimum-scale=1, initial-scale=1, shrink-to-fit=no">
 <title>{% block meta_title %}pan-European news, features and opinions{% endblock %} – Cafebabel</title>
 <meta name=description content="{% block meta_description %}{% endblock %}">
+<!--Social-->
+<meta property="og:url" content="{{ request.url }}">
+<meta property="og:title" content="{{ self.meta_title() }}">
+<meta property="og:image" content="{% block meta_image_absolute_url %}{% endblock %}">
```
davidbgk: There should be a default image, useful when a link to the homepage is shared. Same for other properties, which makes me think if it's pertinent to add meta for pages like edition and so on. Maybe being more specific than the base template is better?

cafebabel/templates/articles/detail.html
```diff
@@ -3,8 +3,12 @@
 
 {% block meta_title %}{{ article.title }}{% endblock %}
 {% block meta_description %}{{ article.summary }}{% endblock %}
+{% block meta_image_absolute_url %}{{ absolute(article.image_url) }}{% endblock %}
+{% block meta_author %}{{ article.author }}{% endblock %}
```
davidbgk: By default, the `article.author` renders the email of the author if he didn't mention a name. That could be spammy! 💣

cafebabel/core/mixins.py
davidbgk: Please remove given that you workaround it.

cafebabel/core/helpers.py
```diff
@@ -59,3 +59,17 @@ def file_exceeds(file_, size):
...
+    domain_lang = {v: k for k, v in lang_domain.items()}
+    return domain_lang.get(request.host)
+
+
+def absolute(url):
+    if not url:
+        return ''
+    if url.startswith('/'):
+        url = url[1:]
+    return f'{request.url_root}{url}'
```
davidbgk, line 75: I found this a bit hackish and prefer the `url_for` way. Not that neither that implementation nor the `url_for` deal with multiple domains like we will have to do (at least to redirect to the appropriate translation)

davidbgk (changes requested):

cafebabel/templates/articles/detail.html
johangiraud: Yep, maybe we shouldn't allow a profile creation without the full name after all.. what do you think ?

cafebabel/templates/base.html
johangiraud: you are right. For exemple pages for edition have no need to be shared with meta

cafebabel/templates/articles/detail.html
davidbgk: Or not use the email as a default name but something engaging? 🤔

cafebabel/templates/base.html
davidbgk: In that case it should not be part of the base template but added as [a macro](http://jinja.pocoo.org/docs/2.10/templates/#macros)?

cafebabel/templates/articles/detail.html
johangiraud: What to you mean engaging ?

cafebabel/templates/base.html
johangiraud: ok, I'll do that

config.py
vinyll: It doesn't override `SERVER_NAME`. It is a mapping for discovering what is the language based on the domain name.

cafebabel/core/helpers.py
vinyll: I feel like there's a very little in common. `url_for` converts a func name into a url. `absolute` converts a relative url into an absolute.
So the url could be a result of `url_for` but is a different process.
How would you see this merged with `url_for`?

davidbgk: `url_for` with `_external=True` parameter gives you an absolute URL (as I previously did).

cafebabel/templates/articles/detail.html
davidbgk: If you call your user `Pinky Lion` or `Crazy Omelette` or whatever it encourages the person to change it right away 😉

cafebabel/templates/base.html
```diff
@@ -1,8 +1,40 @@
+{% from "_meta-macro.html" import meta with context %}
```
davidbgk: I don't think you need the `context` here?

cafebabel/templates/base.html
```diff
@@ -1,8 +1,40 @@
...
 <!doctype html>
 <meta charset=utf-8>
 <meta name=viewport content="width=device-width, minimum-scale=1, initial-scale=1, shrink-to-fit=no">
 <title>{% block meta_title %}pan-European news, features and opinions{% endblock %} – Cafebabel</title>
-<meta name=description content="{% block meta_description %}{% endblock %}">
+
+{% set description -%}
```
davidbgk: Trying to put it within the base template adds extra complexity. I don't get why we try to put meta globally instead of per template/part?

cafebabel/core/helpers.py
vinyll: How would you suggest managing the `article.image_url` in that case?
Should we get back with a method that return the url with the `_external=True` option?

cafebabel/templates/base.html
pointbar: fixed

cafebabel/templates/base.html
pointbar: @davidbgk - the code 'll looked a bit like this?

cafebabel/templates/base.html
```diff
@@ -1,8 +1,28 @@
+{% from "_meta-macro.html" import meta without context %}
 <!doctype html>
 <meta charset=utf-8>
 <meta name=viewport content="width=device-width, minimum-scale=1, initial-scale=1, shrink-to-fit=no">
 <title>{% block meta_title %}pan-European news, features and opinions{% endblock %} – Cafebabel</title>
-<meta name=description content="{% block meta_description %}{% endblock %}">
+
+{% block metas %}
+  {{ meta('name=description', 'pan-European news, features and opinions') }}
```
davidbgk: @pointbar last thing I would change is to only initiate blocks in `base` and put all metas from here to `home.html` template

pointbar: fixed

cafebabel/core/helpers.py
davidbgk: We should have a [dedicated view](http://flask.pocoo.org/docs/0.12/patterns/fileuploads/) for `uploads` (vs. the default `static`) and use it from templates directly with the `_external` parameter, WDYT?

cafebabel/templates/articles/detail.html
```diff
@@ -1,10 +1,31 @@
...
+{% from "_meta-macro.html" import meta without context %}
 {% import "articles/macros.html" as macros without context %}
 
 {% block meta_title %}{{ article.title }}{% endblock %}
-{% block meta_description %}{{ article.summary }}{% endblock %}
+{% block metas %}
+  {{ meta('name=description', article.summary) }}
+  {{ meta('name=author', article.author) }}
+{% endblock %}
+<!--Social-->
```
davidbgk: Not sure that comment is still relevant given how explicit is the block name 👍

cafebabel/templates/base.html
```diff
@@ -1,8 +1,13 @@
+{% from "_meta-macro.html" import meta without context %}
 <!doctype html>
 <meta charset=utf-8>
 <meta name=viewport content="width=device-width, minimum-scale=1, initial-scale=1, shrink-to-fit=no">
 <title>{% block meta_title %}pan-European news, features and opinions{% endblock %} – Cafebabel</title>
-<meta name=description content="{% block meta_description %}{% endblock %}">
+
+{% block metas %}{% endblock %}
+
+<!--Social-->
```
davidbgk: Same here.

cafebabel/templates/home.html
```diff
@@ -1,3 +1,23 @@
 {% extends "base.html" %}
 
+{% block metas %}
+  {{ meta('name=description', 'pan-European news, features and opinions') }}
+  {{ meta('name=author', 'Café Babel') }}
+{% endblock %}
+
+<!--Social-->
```
davidbgk: And there.

cafebabel/core/helpers.py
```diff
@@ -59,3 +59,17 @@ def file_exceeds(file_, size):
     blob = file_.read()
     file_.seek(0)
     return len(blob) > size
+
+
+def current_language():
+    lang_domain = current_app.config.get('DOMAIN_LANGUAGES')
+    domain_lang = {v: k for k, v in lang_domain.items()}
```
davidbgk, line 66: Given that we only use it the reversed way, why don't declaring it directly with `{'domain': 'lang'}` instead?

davidbgk (changes requested):

cafebabel/templates/base.html
pointbar: Fixed

davidbgk: That one has actually been left :D (intentionally?)

cafebabel/templates/articles/detail.html
johangiraud: fixed

davidbgk (approved):

pointbar:
Note:
https://socialmetatags.com/?url=https%3A%2F%2Fpreprod.cafebabel.fr%2F&reload=true

PR #153, opened 2018-01-17
US22 - Tags administration, fix #30

### LINKED ISSUE #30: #US22 - Administrer une page tag
ETQ éditrice je veux gérer une page tag afin que les visiteurs puissent consulter les articles par thèmes

- [ ] - doit permettre d'ajouter une image et/ou une description (markdown) pour personnaliser une page tag
- [ ] - doit permettre de modifier un tag

davidbgk:
ETQ éditrice je veux gérer une page tag afin que les visiteurs puissent consulter les articles par thèmes

- [x] - doit permettre d'ajouter une image et/ou une description (markdown) pour personnaliser une page tag
- [x] - doit permettre de modifier un tag


cafebabel/articles/tags/models.py
```diff
@@ -39,6 +39,18 @@ def get_images_url(self):
     def get_images_path(self):
         return current_app.config.get('TAGS_IMAGES_PATH')
 
+    def save_from_request(self, request):
```
vinyll, line 44: I know we have this for article too and I don't feel so confident about this in the model; it looks more like a view thing.
Let's keep this for now and brain process it for later ;)

cafebabel/articles/tags/views.py
```diff
@@ -36,3 +37,16 @@ def detail(slug):
...
+
+
+@tags.route('/<slug>/edit/', methods=['get', 'post'])
+@editor_required
+def edit(slug):
+    tag = Tag.objects.get_or_404(slug=slug)
+
+    if request.method == 'POST':
+        tag.save_from_request(request)
+        flash('Your tag was successfully saved.')
```
vinyll: This reminds me we are not dealing with errors

cafebabel/templates/articles/tags/edit.html
```diff
@@ -0,0 +1,41 @@
+{% extends "base.html" %}
+
```
vinyll, line 2: We may want to add some basic metas (title only?)

cafebabel/articles/tags/views.py
```diff
@@ -36,3 +37,16 @@ def detail(slug):
     articles = Article.objects(tags__in=[tag], status='published')
     return render_template('articles/tags/detail.html', tag=tag,
                            articles=articles)
+
+
+@tags.route('/<slug>/edit/', methods=['get', 'post'])
+@editor_required
+def edit(slug):
+    tag = Tag.objects.get_or_404(slug=slug)
```
vinyll: Won't it break if we have the same slug in 2 languages? ie: _meetmyhood_ that will remain the same throughout languages

cafebabel/articles/tags/views.py
```diff
@@ -36,3 +37,16 @@ def detail(slug):
     articles = Article.objects(tags__in=[tag], status='published')
     return render_template('articles/tags/detail.html', tag=tag,
                            articles=articles)
+
+
+@tags.route('/<slug>/edit/', methods=['get', 'post'])
+@editor_required
+def edit(slug):
+    tag = Tag.objects.get_or_404(slug=slug, language=current_language())
```
vinyll, line 45: I'd like to attract your attention on this @davidbgk.
This solves the issue when there are [twice the same slug in a different language](https://github.com/cafebabel/cafebabel.com/pull/153/files#diff-0896e0b2368aabe52dcdedab894926a3R106).

However this means that changing the language of the current slug would not make it available when redirecting after save.
Nevertheless, I understood that the expected behavior was to browse/work in the language version that you would be reading/editing.

It may be interesting to exchange about this.

cafebabel/articles/tags/views.py
johangiraud: domain name (.fr, .en, .de..) will do so it's not the same slug, won't it ?

vinyll: Indeed, to select the slug based on the domain we need to filter it out. That's what I [suggested in this commit](https://github.com/cafebabel/cafebabel.com/pull/153/files#diff-a0088a31e32d86af81efe546d99ad4abR45).

cafebabel/static/css/styles.css
```diff
@@ -1454,6 +1454,53 @@ textarea#body {
...
+
+@media screen and (min-width: 750px) {
+  .tag-page header {
+    max-width: 75rem;
+    margin: 30vh auto 2rem auto;
+  }
+}
+
+.tag-page .tag-header-image + header {
+  /*margin-top: 60vh;*/
```
vinyll: delete this

cafebabel/static/js/parallax.js
```diff
@@ -0,0 +1,12 @@
+const parallax = document.querySelector('.tag-header-image')
+window.addEventListener('scroll', () => {
+  const scrolledHeight = window.pageYOffset
+  const limit = parallax.offsetTop + parallax.offsetHeight
+  console.log('scrolledHeight: ', scrolledHeight, 'limit: ', limit)
```
vinyll: remove this

cafebabel/static/js/parallax.js
```diff
@@ -0,0 +1,12 @@
...
+  if (scrolledHeight > parallax.offsetTop && scrolledHeight <= limit) {
+    parallax.style.backgroundPositionY =
+      (parallax.offsetTop - scrolledHeight) / 3 + 'px'
```
vinyll: Template literal?
```js
parallax.style.backgroundPositionY = `${(parallax.offsetTop - scrolledHeight) / 3}px`
```

cafebabel/articles/tags/views.py
johangiraud: I am a bit lost about it now, let's talk about it later @vinyll

cafebabel/templates/articles/tags/detail.html
```diff
@@ -1,25 +1,29 @@
 {% extends "base.html" %}
 
+{% block extra_js %}<script src={{ url_for('static', filename='js/parallax.js') }}></script>{% endblock %}
```
vinyll: this parallax is very specific to tags. It may be interesting to rename it accordingly (_tag-parallax.js_?)

cafebabel/static/js/parallax.js
johangiraud: oh yeah !

cafebabel/templates/articles/tags/detail.html
johangiraud: hum ok !

cafebabel/articles/tags/models.py
davidbgk: Agreed, it's very fragile and we have to change it during the next sprint. Maybe using WTForms?

cafebabel/templates/articles/tags/edit.html
davidbgk: We don't need metas on edit pages.

cafebabel/articles/tags/views.py
davidbgk: If I understand correctly [latest comment](https://github.com/cafebabel/cafebabel.com/issues/30#issuecomment-358245145) from @cafebabel-webmaster we should not be able to change the name, slug of language for a given tag anyway.

cafebabel/templates/articles/tags/detail.html
```diff
@@ -1,20 +1,29 @@
...
+
 {% block meta_title %}{{ tag.name }}{% endblock %}
 {% block meta_description %}{{ tag.summary }}{% endblock %}
 
-{% block body_class %}article-tag{% endblock %}
+{% block body_class %}tag-page{% endblock %}
 {% block body %}
   {% if tag.has_image %}
-    <img src="{{ tag.image_url }}" alt="Image for {{ tag.name }}">
+    <div class=tag-header-image style="background:url({{ tag.image_url }}) no-repeat center 0 fixed"></div>
```
davidbgk: I'm strongly against that change :)

My take is: do whatever you want with JS but keep at least native HTML elements please.

johangiraud: ok, @pointbar sgonna show me how to do else with js, no prob 👌

cafebabel/articles/tags/views.py
davidbgk: OK, now I understand the problem. Thanks for the fix.

cafebabel/articles/tags/models.py
vinyll: [Issue open](https://github.com/cafebabel/cafebabel.com/issues/157)

cafebabel/templates/articles/tags/edit.html
vinyll: I know we don't need SEO but the title might be an indication for history and title bar.
Not big priority though I guess.

davidbgk: Same apply to all edit pages, transversal issue?

vinyll: I suppose it is transverse, but as we are working in // on the meta data it may be why it pops out now.
Not a blocker though.

davidbgk: There are already way too much things to fix within that sprint.

vinyll (approved):

PR #174, opened 2018-01-19
Turn footer date to 2018

davidbgk:


cafebabel/templates/base.html
```diff
@@ -79,7 +79,7 @@
 </main>
 
 <footer>
-  <p>© Cafébabel — 2017</p>
+  <p>© Cafébabel — 2018</p>
```
johangiraud: Possible to have dynamique date ? https://stackoverflow.com/questions/30104914/how-to-get-current-time-in-jinja

johangiraud (changes requested):

cafebabel/templates/base.html
davidbgk: Heh, I knew somebody would come with that, I suspected @vinyll though 😉 

My take is that it consumes resources all the year vs a static 1-char change once a year.

Maybe running the tests for that PR (and subsequent) consumed more than that dynamic date. 🙃 

I let you arbitrate.

johangiraud: Ok ! In the other hand if it's dynamic it prevents us from loading the project in local, make a PR, validate the PR, merge it, deploy it. For you it's easy, for me as well, but will it be for the next _me_?

davidbgk: As I said, it's up to you, it might be indeed more future-proof to make it dynamic.

johangiraud: Bon.. How would you write it dynamically ? I don't get this answer https://stackoverflow.com/a/30252044

johangiraud: Would you say the same about consuming resources if it was in php (I guess `echo date("Y");`). ?

davidbgk:
@cafebabel-webmaster I made a dynamic implementation.

johangiraud:
Supergreat @davidbgk I think it's best this way. Thanks !

johangiraud (approved):

PR #152, opened 2018-01-16
US26 - Page detail for tags, fix #126

davidbgk:
ETQ lecteur, je veux consulter les articles pour un tag donné

- [x] - doit permettre de visualiser la liste des articles pour un tag donné (par ordre antéchrono) : photo de couverture, titres
- [x] - ne doit renvoyer que les articles dans la langue de l'interface
- [x] - peut afficher la photo et/ou la description du tag
- [x] - doit comporter une url partageable de type _/tag/meetmyhood_

PR #178, opened 2018-01-22
display preview article cover when selected or dropped

johangiraud:


cafebabel/static/js/tag-detail.js
```diff
@@ -2,34 +2,22 @@ const container = document.querySelector('.tag-header')
...
     const scrolledHeight = window.pageYOffset
-    const top = container.offsetTop - height * 2
-    container.style.backgroundPositionY = `${(top - scrolledHeight) / 3}px`
+    container.style.backgroundPositionY = 50 + 0.2 * scrolledHeight + '%'
   })
 }
 
 function addBackgroundImage(container, imgUrl) {
-  if (imgUrl) {
+  if (imgUrl === 'None') {
```
davidbgk: You should not have to compare to `None` which is a python keyword, if you have that value within the template please submit a bug or turn it explicitly to an empty string to check the existence here instead of comparing to that string.

johangiraud: That's exactly why I compared `imgUrl` with `None`. Cause I couldn't get it with `if (!value)` or something like that. How can I force `value=""`?

davidbgk: In the template, adding `{{ article.property or '' }}`

johangiraud: done.

cafebabel/templates/articles/tags/detail.html
```diff
@@ -31,7 +31,7 @@
 
 {% block body_class %}tag-page{% endblock %}
 {% block body %}
-  <div class=tag-header data-image={{ tag.image_url }}></div>
+  <div class=tag-header data-image={{ tag.image_url or '' }}></div>
```
davidbgk: Can we switch to a regular `` tag? Then [enhanced](https://en.wikipedia.org/wiki/Progressive_enhancement) by JavaScript?

johangiraud: Why not, but I don't know how to do the parallax effect then. And I already spent too much time on this. Plus, after reading you wikipedia link I still don't really get what you mean by _Then enhanced by JavaScript_

cafebabel/static/css/styles.css
```diff
@@ -845,41 +845,63 @@ button.follow.active:after {
...
   position: relative;
   width: 100%;
-  height: 30vw;
+  height: auto;
   background: transparent;
   transition: all 0.3s;
 }
 
-.create-article-page canvas {
+@media screen and (min-width: 1200px) {
```
vinyll: Are you sure about 1200px for adapting the behavior?

cafebabel/static/css/styles.css
```diff
@@ -845,41 +845,63 @@ button.follow.active:after {
...
+  .create-article-page canvas,
```
vinyll: _canvas_ here is so generic we can't guess what it means unless digging exhaustively into the HTML.
It seems like `.file canvas` would suit – which would be a little more specific and explicit – but is it what you meant?

cafebabel/static/js/image-preview.js
```diff
@@ -0,0 +1,45 @@
...
+const fileInput = document.querySelector('.file input[type="file"]')
+const outputImage = document.querySelector('.file output')
+const sizeInfo = fileInput.parentElement.querySelector('small')
+
+fileInput &&
+  fileInput.addEventListener('change', event => {
+    const image = event.target.files[0]
+    const fileName = image.name
+    const reader = new FileReader()
+    const canvas = document.querySelector('canvas')
```
vinyll: Is this the same `canvas` than the one where you have the noise effect on purpose or is it conflicting?

cafebabel/static/css/styles.css
```diff
@@ -845,41 +845,63 @@ button.follow.active:after {
...
+  .create-article-page .file {
+    width: 120rem;
+    height: 46.7rem;
+  }
+}
+
+.file canvas {
```
vinyll: I see you are using a very different selector than `.create-article-page canvas`. I therefore suppose this is not specific to _create-article-page_ and it should only apply to canvas within _.file_ while the other should apply to any canvas on the page.
🤔

cafebabel/static/css/styles.css
```diff
@@ -1358,6 +1379,17 @@ h1.edit input {
...
+  margin-left: 50%;
+  transform: translateX(-50%);
+}
+
+.article-page a.button:first-of-type {
+  margin: 2rem auto -2rem auto;
+  max-width: 100%;
+}
+
 .article-page article h1 + p {
```
vinyll, line 1418: This is a nice self-explanatory selector BTW 👍

cafebabel/static/js/image-preview.js
vinyll: After re-reading the review I suppose you are trying to point the `canvas` that is inside the `.file`. In which case you could use a selector such as `document.querySelector('.file canvas')`

cafebabel/static/js/image-preview.js
```diff
@@ -0,0 +1,45 @@
+/* display picture file name after selection on profile edit https://tympanus.net/codrops/2015/09/15/styling-customizing-file-inputs-smart-way/ */
+const fileInput = document.querySelector('.file input[type="file"]')
+const outputImage = document.querySelector('.file output')
+const sizeInfo = fileInput.parentElement.querySelector('small')
```
vinyll: Why not selecting `document.querySelector('.file small')` as you did for `outputImage`?

cafebabel/static/js/tag-detail.js
```diff
@@ -2,34 +2,22 @@ const container = document.querySelector('.tag-header')
...
+function parallax() {
   const container = document.querySelector('.tag-header.image')
-  const height = getHeight(imgUrl)
-
   if (!container) return
   window.addEventListener('scroll', () => {
     const scrolledHeight = window.pageYOffset
-    const top = container.offsetTop - height * 2
-    container.style.backgroundPositionY = `${(top - scrolledHeight) / 3}px`
+    container.style.backgroundPositionY = 50 + 0.2 * scrolledHeight + '%'
```
vinyll: template literal ftw: ```… = `${50 + 0.2 * scrolledHeight}%` ```

cafebabel/templates/_macro-extra-js.html
```diff
@@ -0,0 +1,3 @@
+{% macro extra_js(file_path) %}
+  <script src={{ file_path }}></script>
+{% endmacro %}
```
vinyll: newline missing Github says.
You should check for the option in you editor ;)

cafebabel/templates/articles/detail.html
```diff
@@ -14,13 +14,13 @@
   {{ meta('property=og:url', request.url) }}
   {{ meta('property=og:title', self.meta_title()) }}
   {{ meta('property=og:type', 'article') }}
-  {{ meta('property=og:description', 'pan-European news, features and opinions') }}
+  {{ meta('property=og:description', article.summary) }}
   {{ meta('property=og:site_name', 'Cafébabel') }}
```
vinyll, line 18: This should be the value by default in the base template. Is it going to change per page?
If not it should just be hardcoded in the base.

cafebabel/templates/articles/detail.html
```diff
@@ -104,6 +104,7 @@ <h1>{{ article.title }}</h1>
       {% endfor %}
     </ul>
   {% endif %}
+  {{ self.edit_button() }}
```
vinyll, line 107: Is the purpose of this to display twice the edit button?

cafebabel/templates/articles/drafts/create.html
```diff
@@ -1,14 +1,20 @@
 {% extends "base.html" %}
 {% import "articles/macros.html" as macros without context %}
+{% from "_macro-extra-js.html" import extra_js without context %}
 
-{% block extra_js %}<script src={{ url_for('static', filename='js/articles-tags.js') }}></script>{% endblock %}
+{% block extra_js %}
+  {{ extra_js(url_for('static', filename='js/noise.js')) }}
```
vinyll: Would it work if you were writing `{{ extra_js('js/noise.js') }}` here and deport the `url_for` within your macro?
If not there isn't much sense to write `{{ extra_js(` instead of `<script src=`

cafebabel/templates/articles/drafts/create.html
```diff
@@ -1,14 +1,20 @@
 {% extends "base.html" %}
 {% import "articles/macros.html" as macros without context %}
+{% from "_macro-extra-js.html" import extra_js without context %}
```
vinyll: I think this `extra_js` could get into the _meta-macros.html file as it is an HTML meta.

However if you had a reason to think it different, we should challenge the filename. Wondering about what convention we are making with the macro names.
We have: 
- /macros.html
- /_forms-macros.html
- /_meta-macros.html

Maybe we could rename this to _extrajs-macros.html

cafebabel/templates/articles/drafts/create.html
vinyll: Another option would be to write `{{ extra_js(['js/noise.js', 'js/articles-tags.js', 'js/image-preview.js']) }}` and have a `for…endfor` in the macro.

cafebabel/templates/articles/drafts/create.html
```diff
@@ -1,14 +1,20 @@
...
 
 {% block body_class %}create-article-page{% endblock %}
 {% block body %}
   <form method=post action={{ url_for('drafts.create') }} enctype=multipart/form-data>
-  <canvas class=noise></canvas>
   <div class=file>
-    <label for=image>Click here to add a cover picture or Drag & drop your pic here <small>Recommended file size: less than 1Mo</small></label>
+    <canvas class=noise></canvas>
+    <output></output>
+    <label for=image>Click here to add a cover picture or Drag & drop your pic here <small>Recommended file size: 1200x467px, less than 1Mo</small></label>
```
vinyll: Wondering how we should round this for the end user. 1200x500px?

cafebabel/templates/articles/drafts/edit.html
```diff
@@ -1,21 +1,30 @@
 {% extends "base.html" %}
 {% import "articles/macros.html" as macros without context %}
+{% from "_macro-extra-js.html" import extra_js without context %}
+
+{% block extra_js %}
+  {{ extra_js(url_for('static', filename='js/noise.js')) }}
+  {{ extra_js(url_for('static', filename='js/articles-tags.js')) }}
+  {{ extra_js(url_for('static', filename='js/image-preview.js')) }}
+{% endblock %}
```
vinyll: This is so similar to articles/edit.html that you may want to refactor and have this template inheriting from articles/edit.html.
Specific elements could be customized through _blocks_ like we did for draft/detail.html

cafebabel/templates/articles/drafts/edit.html
```diff
@@ -1,21 +1,30 @@
...
+      <output><img src={{ article.image_url }} alt="Image for {{ article.title }}"></output>
+      <div class=checkbox>
+        <label for=delete-image>Delete image</label>
+        <input type=checkbox name=delete-image id=delete-image value=1>
+      </div>
+    {% else %}
+      <canvas class=noise></canvas>
+      <output></output>
+    {% endif %}
+    <label for=image>Click here to add a cover picture or Drag & drop your pic here <small>Recommended file size: 1200x467px, less than 1Mo</small></label>
```
vinyll: This copy-paste comforts the idea of refactoring

cafebabel/templates/articles/tags/detail.html
vinyll: I get what you were talking about now. Indeed having a `style="background-image: {{ tag.image_url }}"` could be one of those rare exception that could make some sense, and then apply the `parralax()` function onto it.

cafebabel/templates/articles/tags/edit.html
```diff
@@ -1,18 +1,27 @@
...
+      <output><img src={{ tag.image_url }} alt="Image for {{ tag.name }}"></output>
       <div class=checkbox>
         <label for=delete-image>Delete image</label>
         <input type=checkbox name=delete-image id=delete-image value=1>
       </div>
+    {% else %}
+      <canvas class=noise></canvas>
+      <output></output>
     {% endif %}
+    <label for=image>Click here to add a cover picture or Drag & drop your pic here <small>Recommended file size: 1200x467px, less than 1Mo</small></label>
```
vinyll: And again.
If for some reason you were not doing the refactoring, it should **at least** be a macro so that you don't forget one text when copy-pasting here and there.

cafebabel/templates/users/edit.html
```diff
@@ -1,7 +1,7 @@
 {% from "users/macros.html" import social_field, default_profile_picture with context %}
 {% extends 'base.html' %}
 
-{% block extra_js %}<script src={{ url_for('static', filename='js/edit-profile.js') }}></script>{% endblock %}
+{% block extra_js %}<script src={{ url_for('static', filename='js/image-preview.js') }}></script>{% endblock %}
```
vinyll: This extra_js block is not using the macro you had written.

cafebabel/static/css/styles.css
johangiraud: yes definitely !

cafebabel/static/css/styles.css
johangiraud: ok, done. You are right

cafebabel/static/css/styles.css
johangiraud: right. Fixed

cafebabel/static/js/image-preview.js
johangiraud: sure.

cafebabel/static/js/image-preview.js
johangiraud: right, need more consistency. Done.

cafebabel/static/css/styles.css
johangiraud: Oh no wait, here was the reason: I have now two different _noise_ effects: one on _create-article-page_ AND one on _tag-page_. That's why I have to precise the one for articles (cause I want a fixed width)

cafebabel/static/js/tag-detail.js
johangiraud: ok done

cafebabel/templates/articles/detail.html
johangiraud: ha in `base.html`?

cafebabel/templates/articles/detail.html
johangiraud: oui ! Made it on purpose

cafebabel/templates/articles/drafts/create.html
johangiraud: good point. I see what you mean when it comes to rename files for consistency. But I don't understand what you mean with: I think this `extra_js` could get into the _meta-macros.html file as it is an HTML meta. ?

johangiraud: I renamed the file.

cafebabel/templates/articles/drafts/create.html
johangiraud: Previous website picture format was 1180x461px. That's why I would like to keep the proportions. What do you think about that @vinyll ?

cafebabel/templates/users/edit.html
johangiraud: ah oui. my bad

johangiraud:
@vinyll about that:

> I get what you were talking about now. Indeed having a `style="background-image: {{ tag.image_url }}"` could be one of those rare exception that could make some sense, and then apply the `parralax()` function onto it.

I first made it this way: 
capture d ecran 2018-01-31 a 10 17 46
(refering to this commit https://github.com/cafebabel/cafebabel.com/pull/152/commits/b12ea62efdfad6ccb83030aab8d960694a5044d3)
and you or @davidbgk said being strongly against that. That's why I changed it and passed it through js with `data-image`. What did change your mind ?

cafebabel/templates/articles/drafts/create.html
vinyll: I meant that `extra_js()` function could also be in the _meta-macro.html file instead of being all alone in its file

cafebabel/templates/articles/drafts/create.html
johangiraud: write `{{ extra_js('js/noise.js') }}` doesn't work :-(

cafebabel/templates/articles/drafts/create.html
johangiraud: ah ok, well it's not a _meta_ per say

cafebabel/templates/articles/drafts/create.html
vinyll: For sure you're right that we should keep the proportions for the existing articles.
My question is more about the dimensions we ask the user to make for new articles; asking the user to make a picture 467px height is so precise that it sounds restrictive. Using a rounded value like 500px (or eventually 450px) could ease that feeling.
But it may be interesting to just quickly ask the editors around instead ;)

vinyll:
https://github.com/cafebabel/cafebabel.com/pull/178#issuecomment-361873001
> and you or @davidbgk said being strongly against that. That's why I changed it and passed it through js with data-image. What did change your mind ?
To make it clear, I'm not **strongly** against it ;)
These are different ways to reach a similar goal and I approve the way you are doing it now with `data-image` attribute. @davidbgk was challenging yesterday the idea that the JS may break or be available in which case the tag image would cease appearing. Possible answers are that it doesn't matter if the image doesn't display and it would still look good enough; other like David was suggesting could be to have the image that would displayed anyways and the parralax would be applied unless the JS would fail loading; or very probably other ideas too 🙂

cafebabel/templates/articles/drafts/create.html
vinyll: The question may sounds silly, but did you plainly try `extra_js('js/noise.js')` with the current code or did you try to adapt the macro function with something like the code below?

```
{% macro extra_js(file_path) %}
  <script src="{{ url_for('static', filename=file_path) }}"></script>
{% endmacro %}
```

cafebabel/templates/users/edit.html
```diff
@@ -1,7 +1,10 @@
-{% from "users/macros.html" import social_field, default_profile_picture with context %}
 {% extends 'base.html' %}
+{% from "users/macros.html" import social_field, default_profile_picture with context %}
+{% from "_extrajs-macros.html" import extra_js without context %}
```
vinyll: BTW, I think writing `{% from "_extrajs-macros.html" import extra_js %}` should do the job and `without context` is optional.

cafebabel/templates/articles/drafts/create.html
vinyll: If it were not working, we could probably keep `{{ extra_js(url_for('static', filename='js/noise.js')) }}` in plain html:`<script src={{url_for('static', filename='js/noise.js')) }}></script>` as we don't gain anything; but I may be missing something here.

cafebabel/templates/articles/macros.html
```diff
@@ -34,3 +34,20 @@
     <ul id=tags-suggestions class=inactive></ul>
   </div>
 {%- endmacro %}
+
+{%- macro image_upload(article) -%}
```
johangiraud: I didn't know how to get `article` dynamic. In tag/edit, I would have liked to use this macros while replacing `article` by `tag`. 
I tried something like:
`<output>Image for {{ {{ article}}.name }}`
but it obviously doesn't work 👎 
any suggestion @vinyll ?

cafebabel/static/css/styles.css
vinyll: Ohh, because of the image width for article that is 1200px I suppose.
Can't we set a max-width/max-height in line 844 instead?
```css
.create-article-page .file canvas,
 .create-article-page .file {
   position: relative;
   width: 100%;
  max-width: 120rem;
  max-height: 46.7rem;
  …
```

cafebabel/templates/articles/detail.html
vinyll: Yes, I suppose in _base.html_

vinyll (commented):
The `.file canvas` things seem more consistent to me :+1:

cafebabel/templates/users/edit.html
johangiraud: ok. Done.

johangiraud:
> Ohh, because of the image width for article that is 1200px I suppose.
Can't we set a max-width/max-height in line 844 instead?
```
.create-article-page .file canvas,
 .create-article-page .file {
   position: relative;
   width: 100%;
  max-width: 120rem;
  max-height: 46.7rem;
```
We dont want max-height to be 46.7rem on mobile :-(

cafebabel/templates/articles/drafts/create.html
johangiraud: Indeed, it would be easier for editors. But if we write a velue like 500px, then we will get images with a 500px height. Won't fit previous format then ? I don't get it.

cafebabel/templates/articles/drafts/create.html
johangiraud: doesn't work like that: `extra_js('js/noise.js')`. It would work with `extra_js('/static/js/noise.js')` though.
Hum you are right, we don't really gain anything here.

cafebabel/static/css/styles.css
vinyll: > We dont want max-height to be 46.7rem on mobile :-(

I'm missing something here. I can't see the `max-height` that is currently applied for mobile and if there's none, I'm not aware of the side effect of a max-height when that height is not reached. Further details will be welcome to me :)

cafebabel/templates/base.html
```diff
@@ -56,15 +56,12 @@
...
-
-<nav>
+<nav id=social-networks>
     <ul>
-        <li><a href=#>Facebook</a></li>
-        <li><a href=#>Twitter</a></li>
-        <li><a href=#>Instagram</a></li>
-        <li><a href=#>LinkedIn</a></li>
-        <li><a href=#>Youtube</a></li>
+        <li><a href=# class=icon-facebook></a></li>
```
vinyll: Still wondering what is the indent size on this project for HTML. Do you go for 2 or 4 spaces?

johangiraud: 2, but I am not going to re-indent manually all the files we have been working on before we set the indent to 2, am I? @vinyll

johangiraud: Possible to add review so that I can merge? @vinyll @davidbgk

cafebabel/templates/base.html
```diff
@@ -79,13 +76,12 @@
 </main>
 
 <footer>
-  <p>© Cafébabel — {{ get_year() }}</p>
+  <p>© Cafébabel — 2017</p>
```
davidbgk: I think you failed a merge here.

johangiraud: ah ok, I fixed it. But there were other chances regarding this chance, right ?

cafebabel/templates/articles/macros.html
```diff
@@ -37,7 +37,7 @@
 
 {%- macro image_upload(object) -%}
   <div class=file>
-    {% if object.image_filename %}
+    {% if object.has_image %}
```
davidbgk: It scares me we have something like that here, as for the `get_date` a merge has probably failed somewhere…

vinyll: Mhhh… I manually typed `has_image` here. Do you mean it should remain `image_filename` in this branch?

davidbgk: No, I mean we should not have a reference to `image_filename` in the template at first. Which commit added it? There is no mention on `master` AFAIK.

vinyll: Well, Im afraid it's messed up then…
screen shot 2018-02-02 at 2 35 43 pm
screen shot 2018-02-02 at 2 35 27 pm

cafebabel/templates/base.html
johangiraud: changes*, change*

cafebabel/static/js/tag-detail.js
```diff
@@ -1,35 +1,20 @@
 const container = document.querySelector('.tag-header')
-const imgUrl = container.getAttribute('data-image')
+const imgUrl = container.querySelector('img')
```
davidbgk: It looks a bit too vague to me, why not `.tag-header img`?

cafebabel/static/js/tag-detail.js
```diff
@@ -1,35 +1,20 @@
...
-
-function getHeight(url) {
-  const img = new Image()
-
-  img.addEventListener('load', () => img.height)
-  img.src = url
-
-  return img.height
+  if (!imgUrl) return
+  container.querySelector('img').classList.add('hidden')
```
davidbgk: Why not reuse `imgUrl`?

cafebabel/templates/articles/tags/detail.html
```diff
@@ -31,7 +31,13 @@
...
 {% block body_class %}tag-page{% endblock %}
 {% block body %}
-  <div class=tag-header data-image={{ tag.image_url }}></div>
+  {% if tag.has_image %}
+    <div class="tag-header image">
+    <img src={{ tag.image_url or '' }} alt="Cover picture for #{{ tag.name }}">
+  {% else %}
+    <div class="tag-header default">
+  {% endif %}
+    </div>
```
davidbgk: From my experience with templating, it's easier on the long run to have if clauses encapsulating the whole `div` for small parts:

```
    {% if tag.has_image %}
      <div class="tag-header image">
        <img src={{ tag.image_url or '' }} alt="Cover picture for #{{ tag.name }}">
      </div>
   {% else %}
      <div class="tag-header default"></div>
   {% endif %}

```

cafebabel/templates/articles/tags/detail.html
```diff
@@ -31,7 +31,13 @@
 
 {% block body_class %}tag-page{% endblock %}
 {% block body %}
-  <div class=tag-header data-image={{ tag.image_url }}></div>
+  {% if tag.has_image %}
+    <div class="tag-header image">
```
davidbgk, line 35: What is the goal of `image` vs. `default`? Don't you want to apply the parallax to the default too?

cafebabel/static/js/tag-detail.js
johangiraud: ok

cafebabel/static/js/tag-detail.js
johangiraud: ok. and I changed `imgUrl` by `bgImg` => more sense, right ?

cafebabel/templates/articles/tags/detail.html
johangiraud: ok

cafebabel/templates/articles/tags/detail.html
johangiraud: when `.image`: `.tag-page .tag-header.image + header{ margin-top:60vh}`
when `.default`: `.tag-page .tag-header.image + header{ margin-top:30vh}`

cafebabel/static/js/tag-detail.js
davidbgk: Even better: `backgroundImage`? :)

PR #186, opened 2018-01-28
Add tests for -articles-tags.js-

pointbar:
@davidbgk & @vinyll - some questions to finalize : 
- I have no idea to the policy for JS libraries installation - ex. npm or /vendors
- I think, it's better if Travis plays the tests, no ?
- For instant, I use JSM, perhaps you prefer Firefox headless or something like Selenium ?

### DIFF
--- a/cafebabel/static/js/articles-tags.js
+++ b/cafebabel/static/js/articles-tags.js
@@ -1,9 +1,12 @@
 class Tags {
   constructor() {
     this.context = document.querySelector('.tags')
-    this.values = Array.from(this.list.querySelectorAll('input')).map(
-      input => input.value
-    )
+  }
+  get tags() {
+    return this.list.querySelectorAll('input')
+  }
+  get tagsNames() {
+    return Array.from(this.tags).map(({ value }) => value)
   }
   get languages() {
     return document.querySelector('#language')
@@ -29,16 +32,14 @@ class Tags {
 
   addNewTag(submission) {
     if (!submission || this._isTag(submission)) return
-    this._addValue(submission)
     this._emptyAddField()
     this._inactiveSuggestionsList()
-    this._render()
+    this._render(this._tagsNamesAdd(submission))
   }
   removeTag(tagElement) {
-    const submission = tagElement.innerText.trim()
+    const submission = tagElement.textContent.trim()
     if (!this._isTag(submission)) return
-    this._removeValue(submission)
-    this._render()
+    this._render(this._tagsNamesRemove(submission))
   }
   handleSuggestion(submission) {
     this._request(submission).then(tagsApi =>
@@ -56,13 +57,13 @@ class Tags {
     )
   }
   _isTag(tag) {
-    return this.values.includes(tag)
+    return this.tagsNames.includes(tag)
   }
-  _addValue(query) {
-    this.values.push(query)
+  _tagsNamesAdd(tagName) {
+    return this.tagsNames.concat([tagName])
   }
-  _removeValue(query) {
-    this.values = this.values.filter(value => value !== query)
+  _tagsNamesRemove(tagName) {
+    return this.tagsNames.filter(selfTagName => selfTagName !== tagName)
   }
   _createSuggestionList(tagsApi) {
     this._activeSuggestionsList()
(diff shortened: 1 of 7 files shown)


cafebabel/static/js/articles-tags.js
```diff
@@ -1,70 +1,82 @@
...
-    return request(
-      `/article/tag/suggest/?language=${this.language}&terms=${submission}`
-    ).catch(console.error.bind(console))
+    const url = `/article/tag/suggest/?language=${
+      this._language
+    }&terms=${submission}`
+    return fetch(url)
+      .then(handleResponse)
+      .catch(error => {
+        const e = new Error(`${error.message} ${url}`)
```
davidbgk: You should not have to do that as it is [already handled](https://github.com/cafebabel/cafebabel.com/blob/master/cafebabel/static/js/utils.js#L8) by the picolib we use, `error.url` should already be there 😉

vinyll:
> I have no idea to the policy for JS libraries installation - ex. npm or /vendors

If you can have one stable final single file for usage, it feels like less of a fuzz than using npm. Unsure about side effects though.

> I think, it's better if Travis plays the tests, no ?

I guess so too. We'll probably _never_ run the tests ourselves.

> For instant, I use JSM, perhaps you prefer Firefox headless or something like Selenium ?

No experience hence no strong opinion.

cafebabel/static/js/articles-tags.js
```diff
@@ -1,42 +1,47 @@
...
-    return languages.options[languages.selectedIndex].value
-  }
-  get list() {
-    return this.context.querySelector('.tags-list')
+  get context() {
+    return document.querySelector('.tags')
   }
   get removeButtons() {
-    return this.list.querySelectorAll('li button')
+    return this._list.querySelectorAll('li button')
```
vinyll: https://gist.github.com/DavidBruant/1016007

cafebabel/static/js/articles-tags.js
```diff
@@ -1,42 +1,47 @@
...
     return this.context.querySelector('input[name=tag-new]')
   }
-  get buttonAdd() {
+  get addButton() {
     return this.context.querySelector('button.add')
   }
   get suggestions() {
+    return this._suggestionsContainer.querySelectorAll('li')
+  }
+  get _languages() {
```
vinyll: Not a big fan of the _wrap-everything_ approach. It does abstract the dom in a good way and would make the code more generic, but it duplicates the lines and the code is not generic in the end anyways.

It does make sense for actual verbe methods (`addNewTag`, `removeTag`) though

cafebabel/templates/articles/macros.html
```diff
@@ -22,8 +22,8 @@
     <ul class=tags-list>
       {% if article and article.tags %}
         {% for tag in article.tags -%}
-          <li class=saved>
-            {{ tag.name }}{# Avoid any whitespace between the name and the input.
+          <li class=saved>{# Avoid any whitespace before the name and the input.
```
vinyll: is it an issue also before?

cafebabel/static/js/articles-tags.js
```diff
@@ -1,42 +1,47 @@
...
+    return Array.from(this._tags).map(({ value }) => value)
+  }
+  get _list() {
+    return this.context.querySelector('.tags-list')
+  }
+  get _suggestionsContainer() {
     return this.context.querySelector('#tags-suggestions')
   }
 
   addNewTag(submission) {
```
vinyll: Why `addNewTag` and not `addTag` while there is `removeTag`?

vinyll (commented):
Just a quick pre-review.
This one will be time demanding for reading either the entire changes at once or understanding per commit.
Not sure how to help 🤔

cafebabel/templates/articles/macros.html
pointbar: I try without, the tests pass and it seems works. So thx

vinyll:
screen shot 2018-02-06 at 5 23 32 pm without even the actual code of the tags and it's usage is triggering an alert to me.
This means we should maintain over 1200 (1500?) lines of code for an input tags field.

Did we ever consider something in Vanilla that is already tested and maintained (like [Choices](https://github.com/jshjohnson/Choices)), [TaggleJS](https://github.com/okcoker/taggle.js) or other libraries?

Not sure how I can help on this. @davidbgk, any thought over all this?

davidbgk:
Reopen when/if we keep the tag custom lib and want to test it.

Issue #190, opened 2018-01-29
US#38 Export data from production related to users


davidbgk:
See https://docs.djangoproject.com/en/2.0/ref/django-admin/#dumpdata

Refs https://github.com/cafebabel/cafebabel.com/issues/189

vinyll:
A dump is available here: `rsync -az prod@91.194.60.65:~/cafebabel/users.json .`

johangiraud:
how should I test this task ?

davidbgk:
@cafebabel-webmaster exceptionally, I'm the tester for that one 😉

johangiraud:
perfect !

PR #198, opened 2018-01-30
Add migration for old users, fix #191

### LINKED ISSUE #191: US#38 Import first/last name and email from prod users
Refs https://github.com/cafebabel/cafebabel.com/issues/189

Dependency https://github.com/cafebabel/cafebabel.com/issues/190

davidbgk:
The migration for 13k users is about 1 minute.

You have to retrieve the data locally first, see #190


.gitignore
```diff
@@ -7,3 +7,4 @@ __pycache__/
 *.egg-info/
 /cafebabel/static/uploads/
 /data/
+*.json
```
vinyll: It feels restrictive for the future to exclude any json file from the project

cafebabel/migrations.py
```diff
@@ -0,0 +1,42 @@
+import json
```
vinyll: Wondering about the name of this file as we will need to deal with regular migrations for database changes in the near future

cafebabel/migrations.py
```diff
@@ -0,0 +1,42 @@
...
+
+def iterate_over_json_file(filename):
+    with open(filename) as json_file:
+        return json.loads(json_file.read())
+
+
+def migrate_users(app, limit):
+    click.echo('Starting users migration.')
+    ds = app.user_datastore
+    old_users = iterate_over_json_file('users.json')
```
vinyll: I assume this file name/path should be the parameter from the command line instead of hardcoded even if it's hopefully a one time script.
That would also fix the gitignore

cafebabel/migrations.py
```diff
@@ -0,0 +1,42 @@
...
+    if limit:
+        old_users = old_users[:limit]
+    bar = ProgressBar(
+        total=len(old_users),
+        template=('{prefix} {animation} {percent} ({done}/{total}) '
+                  'ETA: {eta:%H:%M}'),
+        done_char='👤')
```
vinyll: Looks cool but Github reads it different ^^

cafebabel/migrations.py
```diff
@@ -0,0 +1,42 @@
...
+        return json.loads(json_file.read())
+
+
+def migrate_users(app, limit):
+    click.echo('Starting users migration.')
+    ds = app.user_datastore
+    old_users = iterate_over_json_file('users.json')
+    if limit:
+        old_users = old_users[:limit]
+    bar = ProgressBar(
```
vinyll: 🍸 ☕️ 🍵 🏂

cafebabel/migrations.py
```diff
@@ -0,0 +1,42 @@
...
+        total=len(old_users),
+        template=('{prefix} {animation} {percent} ({done}/{total}) '
+                  'ETA: {eta:%H:%M}'),
+        done_char='👤')
+    editor_role = Role.objects.get(name='editor')
+    with app.app_context():
+        for old_user in bar.iter(old_users):
+            data = old_user['fields']
+            # TODO: deal properly with passwords import, see #192
+            password = data['password']
```
vinyll: I believe this should be just fine as such 👍

cafebabel/migrations.py
```diff
@@ -0,0 +1,42 @@
...
+            creation_date = datetime.strptime(data['date_joined'],
+                                              '%Y-%m-%dT%H:%M:%S')
+            creation_date.replace(tzinfo=timezone.utc)
+            user = ds.create_user(email=data['email'],
+                                  password=password,
+                                  creation_date=creation_date)
+            user.profile.name = f'{data["first_name"]} {data["last_name"]}'
+            user.profile.old_pk = old_user['pk']
+            user.save()
+            if data['is_superuser'] or data['is_staff']:
```
vinyll: oh 🆒 !

vinyll (commented):
I suppose this would be a very appropriate subject for feeding docs/architecture/decisions

cafebabel/migrations.py
davidbgk: What do you mean? The user-kind-emoji is intended!

cafebabel/migrations.py
davidbgk: Alright, I was hesitant between convention vs. configuration for a one-timer

cafebabel/migrations.py
davidbgk: I spent quite some time trying to name it for that reason, any suggestion appreciated 😅

.gitignore
davidbgk: Maybe we should have an `imports` folder to store that kind of files?

cafebabel/migrations.py
vinyll: What about an explicit name like `django-migrations.py`?

vinyll (dismissed):

cafebabel/migrations.py
davidbgk: Good idea 👍

Issue #192, opened 2018-01-29
US#38 Import passwords from old to new accounts


davidbgk:
Refs https://github.com/cafebabel/cafebabel.com/issues/189

In case of impossibility, fallback on asking for a reset password for all users.

vinyll:
Done in #199

PR #199, opened 2018-01-30
a user can also login with a Django legacy password

vinyll:
Solves #192

### DIFF
--- a/cafebabel/users/forms.py
+++ b/cafebabel/users/forms.py
@@ -1,0 +1,50 @@
+from flask_security.forms import LoginForm
+from flask_security.utils import (_datastore, _pwd_context, get_message,
+                                  use_double_hash, hash_password, get_hmac)
+from flask_security.confirmable import requires_confirmation
+
+
+def verify_and_update_password(password, user):
+    """
+    Copy-pasted function from flask_security and disable double_hash for
+    Django passwords.
+    """
+    if (use_double_hash(user.password)
+            and not user.password.startswith('pbkdf2_sha256')):
+        verified = _pwd_context.verify(get_hmac(password), user.password)
+    else:
+        # Try with original password.
+        verified = _pwd_context.verify(password, user.password)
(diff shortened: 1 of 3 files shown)


cafebabel/users/forms.py
```diff
@@ -0,0 +1,50 @@
...
+        Copy-pasted from flask_security and use the altered version of
+        `verify_and_update_password`.
+        """
+        if not super(LoginForm, self).validate():
+            return False
+
+        self.user = _datastore.get_user(self.email.data)
+
+        if self.user is None:
+            self.email.errors.append(get_message('USER_DOES_NOT_EXIST')[0])
```
davidbgk, line 36: How will we deal with translations and `get_message`?

cafebabel/users/forms.py
```diff
@@ -0,0 +1,50 @@
...
+    """
+    if (use_double_hash(user.password)
+            and not user.password.startswith('pbkdf2_sha256')):
+        verified = _pwd_context.verify(get_hmac(password), user.password)
+    else:
+        # Try with original password.
+        verified = _pwd_context.verify(password, user.password)
+    if verified and _pwd_context.needs_update(user.password):
+        user.password = hash_password(password)
+        _datastore.put(user)
```
davidbgk, line 20: I'm (happily) surprised that you can access the datastore directly from flask-security, I thought you would have to retrieve it from the app.

davidbgk (approved):
Really amazed by what you did mate! 😎

cafebabel/users/forms.py
vinyll: This is at least how flask-security deals with it

cafebabel/users/forms.py
vinyll: I haven't taken care of that part and I'm not sure yet.
Dealing with translations for here will be the same way we'll deal with them for flask-security and probably other 3rd party libs

Issue #203, opened 2018-01-31
Preprod is not reloading properly


vinyll:

vinyll:
solved in #204

PR #204, opened 2018-01-31
migrating preprod to gunicorn

vinyll:
#203

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -25,7 +25,7 @@ deploy:
 	${remote} "${goto_src} && git fetch origin ${branch} && git checkout ${branch} && git reset --hard FETCH_HEAD"
 	${remote} "${goto_src} && pip install -r requirements.txt"
 	${remote} "${goto_src} && pkill gunicorn; \
-		gunicorn --daemon -b 0.0.0.0:5000 prod:app \
+		gunicorn --daemon -b 127.0.0.1:5000 prod:app \
 		--error-logfile ~/log/preprod.log --access-logfile ~/log/preprod.log"
 	@echo "> App is deployed. Run \`make logs\` to follow activity."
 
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -14,14 +14,20 @@ help:
...
+	@echo "Log file emptied!"
 
 deploy:
 	@echo "> Fetching master branch and updating sources."
 	${remote} "${goto_src} && git fetch origin ${branch} && git checkout ${branch} && git reset --hard FETCH_HEAD"
 	${remote} "${goto_src} && pip install -r requirements.txt"
-	${remote} "pkill flask && source ~/cafebabel.com/venv/bin/activate && FLASK_APP=prod flask run >> ~/log/preprod-$(date +%Y%m%d).log 2>&1 &"
+	${remote} "${goto_src} && pkill gunicorn; \
+		gunicorn --daemon -b 0.0.0.0:5000 prod:app \
+		--error-logfile ~/log/preprod.log --access-logfile ~/log/preprod.log"
```
davidbgk: Don't you want to have `preprod-access.log` and `preprod-error.log`?

Makefile
```diff
@@ -14,14 +14,20 @@ help:
 	@echo "Make tasks for installation and deployment. Refer to the Makefile content to know more."
 
 logs:  # `make type=access` or `make type=errors` for displaying logs.
```
davidbgk: Either update the comment or better allow to have both (see previous comment)

Makefile
vinyll: I logged it all in one single file as we use this for live viewing errors and there's not much visit here.
If you see any point to get it splitted I can make the adapting @davidbgk.

davidbgk (approved):

PR #210, opened 2018-02-02
Migration script for old articles, fix #195

### LINKED ISSUE #195: US#36 Import articles from dump(s) in models
Refs https://github.com/cafebabel/cafebabel.com/issues/182

davidbgk:
Note: it drops fixtures for articles/users so you will not be able anymore to login with @example.com accounts.


cafebabel/articles/models.py
```diff
@@ -25,11 +25,15 @@ class Article(db.Document, UploadableImageMixin):
     creation_date = db.DateTimeField(default=datetime.datetime.utcnow)
     publication_date = db.DateTimeField()
     tags = db.ListField(db.ReferenceField(Tag, reverse_delete_rule=PULL))
+    old_pk = db.IntField()  # In use for migrations (references in related).
+    old_category_slug = db.StringField()  # In use for redirections.
+    old_slug = db.StringField()  # In use for redirections.
```
vinyll: create an `archive` property with these 3 fields inside?
Could even be an _EmbeddedDocument_ :)

cafebabel/articles/models.py
```diff
@@ -25,11 +25,15 @@ class Article(db.Document, UploadableImageMixin):
...
 
     _translations = None
 
     meta = {
-        'allow_inheritance': True
+        'allow_inheritance': True,
+        'indexes': ['old_pk']
```
vinyll: In which case we'd have `'indexes': ['achive.pk']`

cafebabel/django_migrations.py
```diff
@@ -1,43 +1,174 @@
...
+from cafebabel.articles.models import Article
+from cafebabel.articles.tags.models import Tag
+from cafebabel.articles.translations.models import Translation
+from cafebabel.users.models import Role, User
+from mongoengine import errors as mongo_errors
 from progressist import ProgressBar
 
+PROGRESSIST_TEMPLATE = ('{prefix} {animation} {percent} ({done}/{total}) '
+                        'ETA: {eta:%H:%M} {elapsed}')
+NOW = datetime.now()
```
vinyll, line 14: Please remind me why not storing `utcnow()` instead?

cafebabel/django_migrations.py
```diff
@@ -1,43 +1,174 @@
...
+    return normalized_status
+
+
+def normalize_language(language):
+    normalized_language = language[:2]
+    if normalized_language == 'ge':
+        normalized_language = 'de'
+    elif normalized_language == 'sp':
+        normalized_language = 'es'
+    return normalized_language
```
vinyll: Bewarn that _pol_ is in the list of possible languages in the old system config.
Not sure if it's used though.

vinyll (commented):
💣

cafebabel/django_migrations.py
davidbgk: That date is not stored, just compared to exclude 3-months old content. I guess we are not about an hour precision 😉

cafebabel/django_migrations.py
```diff
@@ -113,32 +117,51 @@ def handle_groups(groups):
...
+    return tags
+
+
+def aggregate_gallery_body(gallery):
+    body = gallery['fields']['body']
+    for image in gallery['images']:
+        image_fields = image['fields']
+        image_src = normalize_image(image_fields['image'])
+        body += f'<p><img src="{image_src}" alt="{image_fields["title"]}"></p>'
+        body += image_fields['body']
```
vinyll: Wondering if we shouldn't got for
```html
<figure>
  <img src="{image_src}" alt="{image_fields["title"]}">
  <figcaption>{image_fields['body']}</figcaption>
</figure>
```

It could be easier to css tune old galleries later.

cafebabel/django_migrations.py
```diff
@@ -113,32 +117,51 @@ def handle_groups(groups):
                 'language': normalize_language(group_fields['language']),
                 'summary': group_fields['about'],
             }
+            if data['language'] == 'xx':
+                # It happens with unused tags(?)
```
vinyll: 🤢

cafebabel/tests/test_articles.py
```diff
@@ -25,6 +25,12 @@ def test_access_article_with_large_slug_should_return_200(client,
     assert response.status_code == HTTPStatus.OK
 
 
+def test_article_with_non_ascii_title(published_article):
```
vinyll: 😎

cafebabel/django_migrations.py
davidbgk: Way better, thanks mate 👍

cafebabel/tests/test_articles_archives.py
```diff
@@ -0,0 +1,32 @@
...
+
+from cafebabel.articles.models import ArticleArchive
+
+
+def test_archive_is_redirect_to_artile(client, published_article):
+    url = 'http://localhost/lifestyle/article/old-artile.html'
+    published_article.archive = ArticleArchive(id=1, url=url)
+    published_article.save()
+    response = client.get(url)
+    assert response.status_code == 301
```
davidbgk: Use HTTPStatus?

cafebabel/django_migrations.py
```diff
@@ -135,8 +141,10 @@ def aggregate_gallery_body(gallery):
     for image in gallery['images']:
         image_fields = image['fields']
         image_src = normalize_image(image_fields['image'])
-        body += f'<p><img src="{image_src}" alt="{image_fields["title"]}"></p>'
-        body += image_fields['body']
+        body += ('<figure>'
+                 f'<img src="{image_src}" alt="{image_fields["title"]}">'
+                 f'<figcaption>{image_fields["body"]}</figcaption>'
+                 '</figure>')
```
vinyll: Other suggestion
```python
body += f'''<figure>
               <img…
         '''
```
More a matter of preference I suppose; I leave this to your consideration

cafebabel/django_migrations.py
```diff
@@ -178,8 +178,8 @@ def create_article(old_article):
     tags = handle_groups(old_article['groups'])
     data = {
         'title': fields['title'],
-        'summary': article_fields['original_header'] or 'TODO',
-        'body': article_fields['body'] or 'TODO',
+        'summary': article_fields['original_header'] or '',
```
vinyll: When a summary was too long, it was ⬇️ moved as the  1️⃣st paragraph into the body and **bolded**.

Shall we ditch this case or treat it?

davidbgk: It's not always the case :/

cafebabel/articles/models.py
```diff
@@ -12,6 +12,12 @@
 from .tags.models import Tag
 
 
+class ArticleArchive(db.EmbeddedDocument):
+    pk = db.IntField()  # In use for migrations (references in related).
+    url = db.StringField()  # In use for redirections.
+    relateds = db.ListField()
```
vinyll, line 18: I see you had put this in the `Article` object in the first place before moving it here.
I shared the opinion of the `Article` property as we will need it the same way for all articles.
What made you change for this?

cafebabel/django_migrations.py
```diff
@@ -1,43 +1,224 @@
...
+        'creation_date': creation_date,
+        'publication_date': timestamp_to_datetime(fields['publication_date']),
+        'author': normalize_author(fields['created_by']),
+        'image_filename': normalize_image(fields['image']),
+        'status': status,
+        'tags': tags or None,
+        'archive': ArticleArchive(
+            pk=old_article['pk'],
+            url=old_article['url'],
+            relateds=fields['relateds']
```
vinyll, line 200: Shouldn't we save the new reference instead?

I was thinking that performance-wise it would be consuming but I think you make 2 rounds anyways.

I suppose if you decide to keep this property in the Archive it makes as much sense to keep the archive _pk_ 🤔

davidbgk: I think we have to make a third round to be really complete and given that we don't need those yet I prefer to store row old ids and to add references to the `Article` when we will implement the feature for all articles.

cafebabel/articles/models.py
davidbgk: See my other comment, here we just store old ids for later use.

cafebabel/core/views.py
```diff
@@ -11,3 +17,11 @@ def home():
 @cores.route('/uploads/<path:filename>')
 def uploads(filename):
     return send_from_directory(current_app.config['UPLOADS_FOLDER'], filename)
+
+
+@cores.route('/<regex("[a-z-]+"):_tag>/<regex("[a-z]+"):_>/<_slug>.html')
```
davidbgk: @vinyll What about turning `<regex("[a-z]+"):_>` into `article`? Are there other cases?

cafebabel/users/models.py
```diff
@@ -21,6 +20,10 @@ class UserProfile(db.EmbeddedDocument, UploadableImageMixin):
     about = db.StringField()
     old_pk = db.IntField()  # In use for migrations (references in articles).
 
+    meta = {
+        'indexes': ['old_pk']
```
vinyll: Should the meta be on the Embedded class or the Article class?
https://stackoverflow.com/a/30455984/328117

vinyll (approved):

vinyll:
🎈

PR #213, opened 2018-02-07
WIP: added language in URL with home redirecting to default by browser

vinyll:
Managing languages is necessary for migrations as articles are contextual to their language.

The website is implementing the language in the URL instead of domain based.

TODO:
- [x] add languages in most urls
- [x] redirect homepage
- [ ] redirect archives


vinyll:
Recreating with a branch name containing the task number

Issue #214, opened 2018-02-07
US#36 Les URL doivent contenir la langue


vinyll:
- [x] add languages in most urls
- [x] redirect homepage
- [x] redirect archives

vinyll:
Done in https://github.com/cafebabel/cafebabel.com/issues/221

PR #215, opened 2018-02-07
add language in URLs

vinyll:
refers to #214

### DIFF
--- a/cafebabel/__init__.py
+++ b/cafebabel/__init__.py
@@ -29,7 +29,7 @@ def register_extensions(app):
     db.init_app(app)
     mail.init_app(app)
 
-    from cafebabel.users.models import Role, User
+    from .users.models import Role, User
 
     app.user_datastore = MongoEngineUserDatastore(db, User, Role)
     security.init_app(app, datastore=app.user_datastore)
@@ -42,12 +42,12 @@ def register_extensions(app):
             pass
 
     if app.testing:
-        from cafebabel.core.testing import ContainsResponse
+        from .core.testing import ContainsResponse
         app.response_class = ContainsResponse
 
 
 def register_blueprints(app):
-    from .core.routing import RegexConverter, LangConverter
+    from .core.routing import LangConverter, RegexConverter
 
     app.url_map.converters['regex'] = RegexConverter
     app.url_map.converters['lang'] = LangConverter
@@ -83,6 +83,7 @@ def register_blueprints(app):
     app.register_blueprint(translations,
                            url_prefix='/<lang:lang>/article/translation')
     app.register_blueprint(users, url_prefix='/<lang:lang>/profile')
+    # Keep that blueprint in the latest position as a fallback.
     app.register_blueprint(archives, url_prefix='')
 
 
@@ -130,7 +131,7 @@ def register_cli(app):
 
 
 def register_template_filters(app):
-    from cafebabel.core import helpers
+    from .core import helpers
 
     app.add_template_filter(quote_plus, 'quote_plus')
     app.add_template_filter(helpers.slugify, 'slugify')
--- a/cafebabel/__init__.py
+++ b/cafebabel/__init__.py
@@ -17,8 +17,8 @@ def create_app(config_object):
     app.config.from_object(config_object)
     app.config.from_pyfile('config.local.py')
 
-    register_extensions(app)
     register_blueprints(app)
+    register_extensions(app)
     register_template_filters(app)
     register_context_processors(app)
     # register_cli is only called when necessary
@@ -47,9 +47,10 @@ def register_extensions(app):
 
 
 def register_blueprints(app):
-    from .core.routing import RegexConverter
+    from .core.routing import RegexConverter, LangConverter
 
     app.url_map.converters['regex'] = RegexConverter
+    app.url_map.converters['lang'] = LangConverter
 
     from .archives.views import archives
     from .articles.views import articles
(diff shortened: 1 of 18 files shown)


cafebabel/__init__.py
```diff
@@ -17,8 +17,8 @@ def create_app(config_object):
     app.config.from_object(config_object)
     app.config.from_pyfile('config.local.py')
 
-    register_extensions(app)
     register_blueprints(app)
+    register_extensions(app)
```
davidbgk: If order matters, add a comment related to that.

cafebabel/__init__.py
```diff
@@ -60,13 +61,28 @@ def register_blueprints(app):
...
+        g.lang = values.pop('lang', app.config['DEFAULT_LANGUAGE'])
+
     app.register_blueprint(cores, url_prefix='')
-    app.register_blueprint(articles, url_prefix='/article')
-    app.register_blueprint(tags, url_prefix='/article/tag')
-    app.register_blueprint(drafts, url_prefix='/article/draft')
-    app.register_blueprint(proposals, url_prefix='/article/proposal')
-    app.register_blueprint(translations, url_prefix='/article/translation')
-    app.register_blueprint(users, url_prefix='/profile')
+    app.register_blueprint(articles, url_prefix='/<lang:lang>/article')
```
davidbgk, line 79: ylang-ylang 🚿

cafebabel/__init__.py
```diff
@@ -60,13 +61,28 @@ def register_blueprints(app):
...
+    app.register_blueprint(tags, url_prefix='/<lang:lang>/article/tag')
+    app.register_blueprint(drafts, url_prefix='/<lang:lang>/article/draft')
+    app.register_blueprint(proposals,
+                           url_prefix='/<lang:lang>/article/proposal')
+    app.register_blueprint(translations,
+                           url_prefix='/<lang:lang>/article/translation')
+    app.register_blueprint(users, url_prefix='/<lang:lang>/profile')
     app.register_blueprint(archives, url_prefix='')
```
davidbgk: We should have added a comment about keeping it last, would you mind adding it?

cafebabel/archives/views.py
```diff
@@ -2,6 +2,8 @@
 
 from flask import Blueprint, redirect, request, url_for
 
+from ..core.helpers import current_language
```
davidbgk: Useless?

cafebabel/__init__.py
```diff
@@ -47,9 +47,10 @@ def register_extensions(app):
 
 
 def register_blueprints(app):
-    from .core.routing import RegexConverter
+    from .core.routing import RegexConverter, LangConverter
```
davidbgk: Switch the two for alphabetical order.

cafebabel/core/helpers.py
```diff
@@ -6,7 +6,7 @@
 from math import ceil
 
 import markdown as markdownlib
-from flask import Markup, abort, current_app, request
+from flask import Markup, abort, current_app, request, g
```
davidbgk: Should come before `request`

cafebabel/core/routing.py
```diff
@@ -5,3 +5,9 @@ class RegexConverter(BaseConverter):
     def __init__(self, url_map, *items):
         super(RegexConverter, self).__init__(url_map)
         self.regex = items[0]
+
+
+class LangConverter(BaseConverter):
+    def __init__(self, url_map, *items):
+        super(LangConverter, self).__init__(url_map)
+        self.regex = '[a-z]{2}'
```
davidbgk, line 13: A bit open, maybe reuse `ALLOWED_LANGUAGES` from django-migrations and make it a setting?

config.py
```diff
@@ -31,9 +31,10 @@ class BaseConfig:
     SECURITY_RECOVERABLE = True
     SECURITY_CHANGEABLE = True
     SECURITY_EMAIL_SENDER = '@'.join(['no-reply', 'cafebabel.com'])
-    SECURITY_POST_LOGIN_VIEW = '/profile/'
+    SECURITY_POST_LOGIN_VIEW = '/en/profile/'
```
davidbgk: Is that required given the `_PREFIX` set later on?

config.py
```diff
@@ -67,24 +68,11 @@ class BaseConfig:
...
 class PreprodConfig(BaseConfig):
-
-    DOMAIN_LANGUAGES = {
-        'fr': 'preprod.cafebabel.fr',
-        'en': 'preprod.cafebabel.co.uk',
-        'de': 'preprod.cafebabel.de',
-        'es': 'preprod.cafebabel.es',
-        'it': 'preprod.cafebabel.it',
-    }
+    pass
```
davidbgk: Just remove it and adapt… wait, we didn't use it 🙃

cafebabel/tests/test_base.py
```diff
@@ -1,8 +1,16 @@
+from http import HTTPStatus
 
-def test_homepage_is_displaying(client):
+
+def test_homepage_is_redirecting_to_default_language(client):
```
davidbgk, line 4: Isn't it quite the same as `test_homepage_is_redirecting_unknown_language_to_default`?

davidbgk:
Kudos for modifying each and every test 😅

cafebabel/core/routing.py
davidbgk: Actually not that easy given that we are not within a context where the `current_app.config` is loaded. Let's keep it that way for now except if you have a better idea :)

cafebabel/tests/test_base.py
vinyll: `test_homepage_is_redirecting_to_default_language` is testing how the homepage will display by default.
`test_homepage_is_redirecting_unknown_language_to_default` will send an unknown _Accepted language_ by the browser to the server and check how the server is handling.

But if it's confusing to you it may also mean that the names are not consistent.
Any suggestion?

davidbgk: It's clearer now, let's go for it 👍

cafebabel/core/routing.py
vinyll: We can't read settings at this stage as we are not yet in the app context

vinyll:
@davidbgk anything you see that is missing?

davidbgk (approved):

PR #220, opened 2018-02-08
Textarea style

johangiraud:


cafebabel/templates/articles/edit.html
```diff
@@ -67,7 +67,7 @@ <h1 class="edit {% if article.id %}active{% endif %}">
     {% endif %}
   </div>
   <div class=textarea>
-    <label for=summary>Summary</label>
+    <label for=summary>Teaser</label>
```
davidbgk, line 70: What about translations and tags which have the same label?

cafebabel/static/js/script.js
```diff
@@ -113,3 +113,38 @@ Array.from(document.querySelectorAll('h1.edit input')).forEach(inputh1 => {
     () => inputh1.value || h1.classList.remove('active')
   )
 })
+
+/* Auto-expand textarea when typing http://jsfiddle.net/hmelenok/WM6Gq/ */
+var textAreaAutoResize
+if (window.attachEvent) {
```
davidbgk: Is that test consistent with other scripts?

cafebabel/templates/articles/edit.html
johangiraud: Translations won't have the same _Teaser_. Never. And for the tags, I think we said we don't care if tags are spelled the same. MeetMtHood in english won't have nothing to do with MeetMyHood in french. They both exists and have a list of articles, that's all. Right ?

cafebabel/static/js/script.js
johangiraud: It works, that's all. I guess it could be cleaner. It that what you mean ?

vinyll: Right, I think we only use `addEventListener` and we don't make backward compatibility to IE5/8

johangiraud: Ah sorry I got what you mean. It could be in a separate script. A script only called on article edit pages

cafebabel/static/js/script.js
```diff
@@ -113,3 +113,38 @@ Array.from(document.querySelectorAll('h1.edit input')).forEach(inputh1 => {
...
+  textAreaAutoResize = function(element, event, handler) {
+    element.attachEvent('on' + event, handler)
+  }
+} else {
+  textAreaAutoResize = function(element, event, handler) {
```
vinyll: As done in other places, `function textAreaAutoresize(…` would be prefered and you could also get rid of `var textAreaAutoResize`

cafebabel/static/js/script.js
davidbgk: No, I mean we did not take care of that retro-compatibility since now.

cafebabel/static/js/script.js
```diff
@@ -113,3 +113,38 @@ Array.from(document.querySelectorAll('h1.edit input')).forEach(inputh1 => {
...
+    element.addEventListener(event, handler, false)
+  }
+}
+function textAreaResizing() {
+  Array.from(document.querySelectorAll('textarea')).forEach(textArea => {
+    function resize() {
+      textArea.style.height = 'auto'
```
vinyll: Why do we first set it to `'auto'` and then immediately change it?

cafebabel/templates/articles/edit.html
davidbgk: For tags I get it but for translations you will keep Summary within the form and it will be inconsistent with that one (Teaser)

cafebabel/static/js/script.js
```diff
@@ -113,3 +113,38 @@ Array.from(document.querySelectorAll('h1.edit input')).forEach(inputh1 => {
...
+    element.attachEvent('on' + event, handler)
+  }
+} else {
+  textAreaAutoResize = function(element, event, handler) {
+    element.addEventListener(event, handler, false)
+  }
+}
+function textAreaResizing() {
+  Array.from(document.querySelectorAll('textarea')).forEach(textArea => {
+    function resize() {
```
vinyll: Not certain why we need to declare this function inside of the forEach 🤔

cafebabel/static/js/script.js
johangiraud: Ok you'll explain that later

cafebabel/templates/articles/edit.html
johangiraud: should I change the `for=summary` to `for=teaser`?

vinyll: I suppose that's a good question that will occur in multiple places. We might need a namespace that would make a context. But that will be when translating I suppose.

cafebabel/static/js/script.js
vinyll: @cafebabel-webmaster https://stackoverflow.com/a/30915025/328117.
We don't treat older browsers cases.

cafebabel/templates/articles/edit.html
davidbgk: @cafebabel-webmaster no, I mean [here](https://github.com/cafebabel/cafebabel.com/blob/master/cafebabel/templates/articles/translations/edit.html#L17) and [there](https://github.com/cafebabel/cafebabel.com/blob/master/cafebabel/templates/articles/translations/create.html#L19) it will create inconsistency in the naming.

cafebabel/static/js/script.js
johangiraud: ok

cafebabel/static/js/script.js
johangiraud: I worked only on the first `<textarea>` otherwise

cafebabel/static/js/script.js
johangiraud: I don't know.

cafebabel/static/js/script.js
```diff
@@ -116,15 +116,11 @@ Array.from(document.querySelectorAll('h1.edit input')).forEach(inputh1 => {
...
-if (window.attachEvent) {
-  textAreaAutoResize = function(element, event, handler) {
-    element.attachEvent('on' + event, handler)
-  }
-} else {
-  textAreaAutoResize = function(element, event, handler) {
-    element.addEventListener(event, handler, false)
-  }
+
+textAreaAutoResize = (element, event, handler) => {
```
vinyll: this should be a `function textAreaAutoResize…` and remove the `var textAreaAutoResize`

cafebabel/static/js/script.js
vinyll: yes, but if you declare the function outside and you pass the `textArea` as an argument.
I suppose there was a reason but it looks very unsusual. Maybe it has to do with the setTimeout. (?)

vinyll (changes requested):

cafebabel/static/js/script.js
vinyll: Can we remove it then?

johangiraud: ok

cafebabel/static/js/script.js
johangiraud: I can't get it work else way :-(

cafebabel/static/js/script.js
johangiraud: ok

vinyll (approved):

cafebabel/static/js/script.js
```diff
@@ -113,3 +113,31 @@ Array.from(document.querySelectorAll('h1.edit input')).forEach(inputh1 => {
     () => inputh1.value || h1.classList.remove('active')
   )
 })
+
+/* Auto-expand textarea when typing http://jsfiddle.net/hmelenok/WM6Gq/ */
```
vinyll: This whole script being an external library, the functions it should be in an isolated file and then called here like you did on line 143.

cafebabel/templates/base.html
```diff
@@ -56,12 +56,15 @@
   </ul>
 </nav>
 
+<button id=participate><a href=#>Participate</a></button>
```
vinyll: Interactive element (`a`) inside of an interactive element (`button`) looks confusing.
It should either be a link to a different content (`a`), or be an interactive element (`button`).

`a href=#` makes me think it's not an actual link…

cafebabel/templates/base.html
```diff
@@ -56,12 +56,15 @@
...
+<button id=translate><a href=#>Translate</a></button>
+
 <nav id=social-networks>
   <ul>
-    <li><a href=# class=icon-facebook></a></li>
-    <li><a href=# class=icon-twitter></a></li>
-    <li><a href=# class=icon-instagram></a></li>
-    <li><a href=# class=icon-youtube></a></li>
+    <li><a href=https://www.facebook.com/cafebabelenglish class=icon-facebook></a></li>
```
vinyll, line 72: :+1: This will permit to test properly and know what the links look like.

vinyll (changes requested):
Sorry, I meant to approve the last commit and I didn't realize it was mixed into this larger PR.

cafebabel/static/js/script.js
johangiraud: ok done. https://github.com/cafebabel/cafebabel.com/pull/220/commits/d149accdc65015f66500169fea80accaed71772f

vinyll:
I suppose it's way too late now, but this PR is suppose to be about textarea styles.
It seems like a lot has got nothing to do with the original point of the PR.

It's very fine as long as we are all ok that it will therefore take more time to review and to merge as one blockage will block it all. ;)

johangiraud:
@vinyll you are totally right. I find sometimes easier to do minor changes in the same PR instead of reopening another one. But that's WRONG !!!

cafebabel/templates/articles/edit.html
vinyll: Ok, in other word here for an article we call it "Teaser", but [for a translation we call it "Summary"](https://github.com/cafebabel/cafebabel.com/blob/master/cafebabel/templates/articles/translations/edit.html#L17).
The question is why do we call it different here, or if they all should be called "Teaser", or this one should remain "Summary".

cafebabel/static/js/textarea-change.js
```diff
@@ -0,0 +1,27 @@
...
+    textAreaAutoResize(textArea, 'cut', delayedResize)
+    textAreaAutoResize(textArea, 'paste', delayedResize)
+    textAreaAutoResize(textArea, 'drop', delayedResize)
+    textAreaAutoResize(textArea, 'keydown', delayedResize)
+
+    resize()
+  })
+}
+
+textAreaResizing()
```
vinyll: I didn't see this one line should be here but I guess it make sense too :)

vinyll (commented):
[This point](https://github.com/cafebabel/cafebabel.com/pull/220/files#r166966400) remains pending.

cafebabel/static/js/textarea-change.js
johangiraud: well since `textAreaResizing` is a function, it doesn't call it itself, does it? So we have to call it at one point

johangiraud:
About the _summary_ issue, I have replaced to one left and changed it to _Teaser_. Anyway since the beginning, you guys have chosen the words in use. When I thought we could translate the website in 5 languages, I planned to go through the whole _wording_ issue is the editors when we would have translated the website. But since we can't do this anymore, I guess, we have to do that for the english version very soon with Lara (english editor).

cafebabel/templates/articles/edit.html
johangiraud: done: https://github.com/cafebabel/cafebabel.com/pull/220/commits/ee76d5d98f6eee4f2af284c9157c206fad47b4a3, https://github.com/cafebabel/cafebabel.com/pull/220/commits/1156dab3d50e84989a5bb2bb13f1975655722fb2

cafebabel/static/js/textarea-change.js
```diff
@@ -0,0 +1,27 @@
+/* Auto-expand textarea when typing http://jsfiddle.net/hmelenok/WM6Gq/ */
+function textAreaAutoResize(element, event, handler) {
```
davidbgk: I don't get why we need that function instead of directly calling `addEventListener` below?

cafebabel/static/js/textarea-change.js
davidbgk: A common pattern in JS is to use an [IIFE](https://en.wikipedia.org/wiki/Immediately-invoked_function_expression)

cafebabel/static/js/textarea-change.js
vinyll: OH, I think it's because in the copied version there was a switch between `addEventListener` and `attachEvent`. I guess there's no more reason now.

cafebabel/templates/articles/translations/detail.html
```diff
@@ -7,26 +7,26 @@
   {{ meta('name=description', translation.summary|striptags) }}
 {% endblock %}
 
+{% block body_class %}article-page{% endblock %}
 {% block body %}
 <article>
-  <h1>{{ translation.title }}</h1>
-
-  <p class=reading-time>Reading time: {{ translation.body|reading_time }} min.</p>
```
davidbgk, line 14: No more reading time for translations?

cafebabel/static/js/textarea-change.js
vinyll: Most of the purpose of IIFEs may have been replaced with the local scope of ES6 (that's one reason of using `const`).
Therefore we could have the code directly outside of a function but I guess it it more readable this way.
I had in mind to call `textAreaResizing()` from the main _scripts.js_ file but it doesn't feel ackward to me as such.
Any thought?

cafebabel/templates/articles/translations/detail.html
johangiraud: no..

cafebabel/static/js/textarea-change.js
johangiraud: Feel free to edit this yourself please @vinyll @davidbgk . I don't know the difference between js versions. Just make it consistant with the rest we've been doing: ES6, right?

cafebabel/static/js/textarea-change.js
davidbgk: It would mean that `textarea-change.js` should be included everywhere `scripts.js` is in use?

cafebabel/static/js/textarea-change.js
johangiraud: I am sure you are right, and you see in 2 sec if we need this function or if we don't

cafebabel/static/js/textarea-change.js
johangiraud: I don't think we need `texarea-change.js` each time we need `script.js`. I am even pretty sure of it.

vinyll (approved):
It seems fine to me

johangiraud:
great !

PR #230, opened 2018-02-09
Revert part of previous changes, refs #228

davidbgk:
Especially the fact to fallback on redirection to preprod matching.

Get rid of SERVER_NAME even for tests.


vinyll (approved):
Ok, nothing fancy here.
Just getting the uri part and finding it thoughout the archives.
Limit case: 2 articles that would have the same uri in a different language which cannot be processed with this method.

PR #247, opened 2018-02-22
allow textarea to resize on 'cut' event or when deleting content

johangiraud:
If we dont put `textArea.style.height = 'auto'` it doesn't resize when deleting content :/

### DIFF
--- a/cafebabel/static/js/textarea-change.js
+++ b/cafebabel/static/js/textarea-change.js
@@ -2,7 +2,7 @@
 Array.from(document.querySelectorAll('textarea')).forEach(textArea => {
   function resize() {
     textArea.style.height =
-      'auto' /* hack to allow textarea to be resiized when deleting or cut content */
+      'auto' /* hack to allow textarea to be resized when deleting or cut content */
     textArea.style.height = `${textArea.scrollHeight}px`
   }
 
--- a/cafebabel/static/js/textarea-change.js
+++ b/cafebabel/static/js/textarea-change.js
@@ -1,7 +1,8 @@
 /* Auto-expand textarea when typing http://jsfiddle.net/hmelenok/WM6Gq/ */
 Array.from(document.querySelectorAll('textarea')).forEach(textArea => {
   function resize() {
-    textArea.style.height = 'auto'
+    textArea.style.height =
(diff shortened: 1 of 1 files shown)


davidbgk:
> it doesn't resize when deleting content

What is the value of `textArea.scrollHeight` in these cases?

johangiraud:
> What is the value of textArea.scrollHeight in these cases?

It stays the same.
Now it can be resize if you delete text. Beside that, we loose the transition when the textarea is resized :-( But i know how to do else.

davidbgk:
> It stays the same.

In that case, do we need to set again `textArea.style.height` after the `auto` attribution?

johangiraud:
> In that case, do we need to set again textArea.style.height after the auto attribution?

Bah oui, otherwise how do you want it to be resized when typing more text?

cafebabel/static/js/textarea-change.js
```diff
@@ -1,6 +1,7 @@
 /* Auto-expand textarea when typing http://jsfiddle.net/hmelenok/WM6Gq/ */
 Array.from(document.querySelectorAll('textarea')).forEach(textArea => {
   function resize() {
+    textArea.style.height = 'auto'
```
vinyll: Because it is unusual to see, I'd rather read a comment in the code that tells _why_ we need to hack it so we're not tempted to clean it off in the next refactoring ;)

johangiraud: ok !

cafebabel/static/js/textarea-change.js
```diff
@@ -1,6 +1,8 @@
 /* Auto-expand textarea when typing http://jsfiddle.net/hmelenok/WM6Gq/ */
 Array.from(document.querySelectorAll('textarea')).forEach(textArea => {
   function resize() {
+    textArea.style.height =
+      'auto' /* hack to allow textarea to be resiized when deleting or cut content */
```
vinyll: _resized_

vinyll (approved):

cafebabel/static/js/textarea-change.js
johangiraud: aie

PR #251, opened 2018-02-22
232 - Fix: can't insert tags properly

### LINKED ISSUE #232: Impossible d'ajouter de nouveau tags lors de la création d'un article
**URL**: draft/create ou draft/edit ou article/edit
**Contexte**: je suis en train d'éditer ou de créer un article et je souhaite ajouter des tags pour mieux classer l'article
**Problème**: impossible d'ajouter des tags
**Hint**: la console me montre :
```
Error
    at fetch.then.catch.error (utils.js:8)
    at <anonymous>
Promise.catch (async)
_request @ articles-tags.js:49
handleSuggestion @ articles-tags.js:42
inputNewTag.addEventListener.event @ articles-tags.js:178
articles-tags.js:68 Uncaught (in promise) TypeError: Cannot read property 'forEach' of undefined
    at Tags._createSuggestionList (articles-tags.js:68)
    at _request.then.tagsApi (articles-tags.js:43)
    at <anonymous>
```

Par contre ça fonctionne lorsqu'on tape un tag dans l'`input` puis qu'on appuie directement sur le bouton _save_. Weird.

### LINKED ISSUE #250: Problèmes avec les tags en plusieurs mots
**Url** : 
Toutes les pages articles en édition
**Contexte** : 
Lorsque qu'un tag en deux mots est enregistré dans un article il bloque la saisie de nouveaux tags

pointbar:
This closes #250, closes #232

### DIFF
--- a/cafebabel/static/js/articles-tags.js
+++ b/cafebabel/static/js/articles-tags.js
@@ -87,7 +87,7 @@ class Tags {
     return this._isTagSaved(tagValue).then(
       isSave =>
         `<li ${isSave ? 'class=saved' : ''}>${tagValue}
-          <input name=tag-${++index} list=tags value=${tagValue} type=hidden>
+          <input name=tag-${++index} list=tags value="${tagValue}" type=hidden>
           <button type=button></button>
         </li>`
     )
--- a/cafebabel/static/js/articles-tags.js
+++ b/cafebabel/static/js/articles-tags.js
@@ -45,7 +45,7 @@ class Tags {
   }
   _request(submission) {
     return request(
-      `/article/tag/suggest/?language=${this.language}&terms=${submission}`
+      `/${this.language}/article/tag/suggest/?language=${this.language}&terms=${submission}`
     ).catch(console.error.bind(console))
   }
   _isTagSaved(submission) {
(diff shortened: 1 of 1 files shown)


cafebabel/static/js/articles-tags.js
```diff
@@ -5,9 +5,11 @@ class Tags {
       input => input.value
     )
   }
+  get languages() {
```
vinyll: `get language()` returns a language string.
An intuitive thought would be that `get languages()` returns a list of languages string, but it's one DOM node that probably contains things about language(s).
I'd suggest we rename it to something like `get languageDom()`? `get languageNode()`? or another name that would reflect more what it actually is.

vinyll: Mhhh… after reading the usage below another option if we want to keep the name `get languages()` could be to return the list of languages:
```js
get languages() {
  return Array.from(document.querySelector('#language').options).map(option => option.value)
}
```
(code is untested)

davidbgk:
I don't get how that PR is fixing https://github.com/cafebabel/cafebabel.com/issues/250 I should miss something obvious.

cafebabel/static/js/articles-tags.js
pointbar: @vinyll - I have to keep the selected language, so I begin to implement your first thought

davidbgk (approved):

PR #260, opened 2018-03-01
implemented authorS to match multiple authors from archives

### LINKED ISSUE #249: Wrong authors for imported articles
Sometimes, we attribute the wrong author to a given article (translation?) when importing old data. It has to be discussed with editors to find pathologic cases.

vinyll:
Fix #249.
Backend implementation for #258.


cafebabel/articles/models.py
```diff
@@ -71,6 +72,16 @@ def is_translation(self):
     def is_translated_in(self, language):
         return bool(self.get_translation(language))
 
+    @property
+    def author(self):
+        return len(self.authors) and self.authors[0]
```
davidbgk: Is the `len()` necessary?

cafebabel/articles/models.py
```diff
@@ -71,6 +72,16 @@ def is_translation(self):
...
+
+    @author.setter
+    def author(self, value):
+        if value in self.authors:
```
davidbgk: A comment/todo would be cool to understand why we are doing this for now.

cafebabel/articles/translations/views.py
```diff
@@ -23,7 +23,7 @@ def create():
                 original_article=article.id,
                 status='draft',
                 editor=article.editor,
-                author=article.author,
+                authors=[article.author],
```
davidbgk: Why not `article.authors`?

cafebabel/articles/models.py
```diff
@@ -92,10 +103,10 @@ def save_from_request(self, request):
         if current_user.has_role('editor'):
             if not self.editor:
                 data['editor'] = current_user.id
-            data['author'] = User.objects.get(id=data['author'])
+            data['authors'] = [User.objects.get(id=data['author'])]
```
davidbgk: I'm afraid it erases previous multiple authors?

cafebabel/django_migrations.py
```diff
@@ -217,8 +217,8 @@ def create_article(old_article):
             # Will be checked again on second pass.
             # click.echo(f'Article does not exist: {old_pk} (skipping)')
             return
-        translator = data['author']
-        data['author'] = original_article.author
+        translator = data['authors'][0]
+        data['authors'] = [original_article.author]
```
davidbgk: Why not `original_article.authors`?

davidbgk (changes requested):
I think there are still missing tests for edge cases.

cafebabel/articles/translations/views.py
vinyll: That makes me wonder what should happen when updating the list of the original article's authors.
I suppose the translations should follow.

vinyll:
> I think there are still missing tests for edge cases.

I guess at a certain point we should decide to implement the functionality on the frontend side and this should avoid most edge-cases.

Seeing the management of the authors for translations makes me wonder if Translation authors shouldn't actually be a virtual property instead reading from `original_article.authors`?

vinyll (commented):
LGTM

davidbgk:
> LGTM

Still importing locally, I'll not approve before ;)

davidbgk:
For the moment, authors for articles are fixed with that PR but not translators which remains the first author (which appears to be wrong)

vinyll:
Indeed, the translations have a `.translators()` methods just like the articles have `.authors()`.
That will need to be implemented…

cafebabel/tests/test_users.py
```diff
@@ -113,3 +113,30 @@ def test_login_complete_redirects_if_not_logged_in(client, user):
...
     assert ('/en/login?next=%2Flogin_complete%2F'
             in response.headers.get('Location'))
+
+
+def test_user_suggest_basics(client, user):
+    response = client.get('/en/profile/suggest/?terms=user@')
+    assert response.status_code == HTTPStatus.OK
+    assert response.json == [{
+        'name': 'user@example.com',
+        'pk': str(user.pk)
```
vinyll, line 123: We shall need to be consistent between `pk` and `id` though the project. We've been mostly using `id`, and seldomly `pk`.

vinyll:
Tests are fixed, @cafebabel-webmaster may want to customize what has been adapted on the frontend.
Appart of that it seems to be all correct…

johangiraud:
Ok I'll work on it now

cafebabel/static/css/styles.css
```diff
@@ -1033,6 +1033,93 @@ h1.edit input {
   text-align: center;
 }
 
+.create-article-page .select:nth-child(2) select {
+  margin-bottom: 5rem;
+}
+
+/* overwrite choices styles */
```
davidbgk: I think these styles could be in their dedicated custom CSS, this way…

cafebabel/templates/articles/edit.html
```diff
@@ -2,8 +2,12 @@
 {% import "articles/macros.html" as macros %}
 {% from "_extrajs-macros.html" import extra_js %}
 
+{% block extra_head %}
+  <link rel=stylesheet href={{ url_for('static', filename='choices/css/choices.min.css') }}>
+{% endblock %}
```
davidbgk: …it will be possible to put them here after the lib one and…

cafebabel/templates/base.html
```diff
@@ -7,6 +7,7 @@
 
 {% block metas_social %}{% endblock %}
 
+{% block extra_head %}{% endblock %}
```
davidbgk: …to re-swap this block below the `style.css` inclusion.

cafebabel/static/css/styles.css
johangiraud: but we can't overwrite their css within their css files, right ? so ?

davidbgk: No, but you can create a new dedicated CSS with custom styles for that component.

johangiraud: 👍

cafebabel/templates/articles/edit.html
```diff
@@ -46,8 +46,8 @@ <h1 class="edit {% if article.id %}active{% endif %}">
         </select>
       </label>
     </div>
-    <div>
-      <label for=authors>Authors
+    <div class=author-selector>
```
vinyll: you may want to give it a more generic class name when it will be reused. Something like "select-multi" or whatever.

cafebabel/templates/base.html
johangiraud: ok ! got it !

cafebabel/templates/articles/macros.html
```diff
@@ -2,19 +2,20 @@
...
-      {% else %}
-        {{ default_profile_picture() }}
-      {% endif %}
-      {{ author }}
-    {% endfor %}
-    <br>
-    Reading time: {{ article.body|reading_time }} min
+    <span>Published on <time pubdate="{{ article.creation_date.date() }}">{{ article.creation_date.date() }}</time></span> | Reading time: {{ article.body|reading_time }} min
   </p>
+    <div class=author-container>
```
vinyll: unindent this block

johangiraud: ok https://github.com/cafebabel/cafebabel.com/pull/260/commits/562cb827bbaefa2fa0e9f4e865cee07f17724622

cafebabel/templates/articles/edit.html
johangiraud: ok https://github.com/cafebabel/cafebabel.com/pull/260/commits/6c8b5fea95ac0f78636f41bc0599309d55ee5056

johangiraud:
in _test_translation.py_ I killed:
```
def test_translation_published_should_have_translator(client, translation):
    translation.status = 'published'
    translation.save()
    translator = translation.translators[0]
    response = client.get(f'/en/article/{translation.slug}-{translation.id}/')
    assert ((f'Translated from '
             f'<a href="/en/article/draft/{translation.original_article.id}/">'
             f'article title') in response)
    assert (f'by <a href="/en/profile/{translator.id}/">{translator}</a>.'
            in response)
```
because I now display the translator right next to the author(s). And I no longer display it at the article bottom. I don't get how to change properly the test. Can you help ?

cafebabel/tests/test_translations.py
```diff
@@ -244,8 +244,10 @@ def test_translation_published_should_have_translator(client, translation):
     assert ((f'Translated from '
              f'<a href="/en/article/draft/{translation.original_article.id}/">'
              f'article title') in response)
-    assert (f'by <a href="/en/profile/{translator.id}/">{translator}</a>.'
+    assert (f'<a href="/en/profile/{translator.id}/" class=translator-link>'
             in response)
+    assert f'{translator}' in response
```
vinyll: That's light by I get the workaround

johangiraud: 👍

johangiraud:
capture d ecran 2018-03-07 a 16 55 37
ok here is the problem. I have juste created a translation and I land on this page. I would like to see displayed the translator profile link right next to the author(s) like that:
capture d ecran 2018-03-07 a 16 59 01
How do I do that?
I tried `{% if article.is_translation or article.is_translations_drafts %}` in _article/macros.html_ but it doesn't work..

davidbgk:
>  How do I do that?

@cafebabel-webmaster it's because we do not call the `article_meta` macro in `translations/detail.html` template, 2 options: either use it or adapt directly the template.

johangiraud:
> @cafebabel-webmaster it's because we do not call the article_meta macro in translations/detail.html template, 2 options: either use it or adapt directly the template.

We do line 3 : `{% import "articles/macros.html" as macros without context %}` @davidbgk 

I can't have it work :-( `{% if article.is_translation or article.is_translation %}`?

davidbgk (approved):
🚀

cafebabel/templates/articles/to-translate.html
```diff
@@ -1,4 +1,5 @@
 {% extends "base.html" %}
+{% import "articles/macros.html" as macros without context %}
```
vinyll: `without context` is implicit; can be removed

PR #261, opened 2018-03-01
Fix duplicate comparison of mongo objects

davidbgk:
We need to compare strings of primary keys because of mongo duality of ObjectIDs vs. raw strings.

### DIFF
--- a/cafebabel/articles/models.py
+++ b/cafebabel/articles/models.py
@@ -51,6 +51,11 @@ class Article(db.Document, UploadableImageMixin):
     def __str__(self):
         return self.title
 
+    def __eq__(self, other):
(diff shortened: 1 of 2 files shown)


cafebabel/articles/models.py
```diff
@@ -51,6 +51,11 @@ class Article(db.Document, UploadableImageMixin):
     def __str__(self):
         return self.title
 
+    def __eq__(self, other):
+        # We need to compare strings of primary keys because of mongo
+        # duality of ObjectIDs vs. raw strings.
+        return str(self.pk) == str(other.pk)
```
vinyll, line 57: 💅

vinyll (approved):

PR #264, opened 2018-03-01
set debug mode for preprod

vinyll:
solves #208


davidbgk (approved):
Even if performances will be worse with that ;)

davidbgk:
@vinyll it's `fix`or `fixes` but `solves` doesn't look to work for auto-closing.

PR #272, opened 2018-03-02
Participation button styles

johangiraud:


cafebabel/static/css/styles.css
```diff
@@ -1591,13 +1624,43 @@ textarea#body {
...
+  color: #00283c;
+  position: static;
+}
+
+.article-language-selector li.original-language a:hover {
+  background-color: #b4db13;
+}
+
+.article-language-selector li.original-language a:before {
+  content: 'Original language';
```
vinyll: This is inaccessible and will be untranslatable; it should be in the HTML.
As the attribute of the `a` tag maybe? (`<a title="Original language"…>`)

cafebabel/static/css/styles.css
```diff
@@ -1610,6 +1673,9 @@ textarea#body {
     margin-left: 0.5rem;
     vertical-align: middle;
   }
+  .article-language-selector li.original-language a:before {
+    display: block;
```
vinyll, line 1677: I'm not sure why you can't write it in https://github.com/cafebabel/cafebabel.com/pull/272/files#diff-6569206075db87419fc4b34e54f5eb4cR1645.
Just a quick reply will help ;)

cafebabel/templates/articles/detail.html
```diff
@@ -54,7 +54,7 @@ <h1>{{ article.title }}</h1>
       <ul>
         <li><span>{{ article.language }}</span></li>
           {% if article.is_translation %}
-            <li class=original-language>Translated from <a href="{{ article.original_article.detail_url }}">{{ article.original_article.language }}</a></li>
+            <li class=original-language><a href="{{ article.original_article.detail_url }}">{{ article.original_article.language }}</a></li>
```
vinyll: If you want to play in CSS with this text you could wrap it with a `span`: `<span>Translated from</span> `
(I suppose that's why you moved it from here to CSS but I may have misunderstood)

cafebabel/static/css/styles.css
johangiraud: Right, done.

cafebabel/static/css/styles.css
johangiraud: I don't want it to be displayed on mobile neither on click nor "always". And if I use `display:none` always and `display:block` on hover, I can't get any smooth transition. Was that what you meant?

cafebabel/templates/articles/detail.html
johangiraud: I thought it was clearer to have _Original language_ than _Translated from_. What do you think?

cafebabel/static/css/styles.css
vinyll: Right, makes sense now 👍

cafebabel/static/css/styles.css
```diff
@@ -1643,7 +1643,7 @@ textarea#body {
 }
 
 .article-language-selector li.original-language a:before {
-  content: 'Original language';
+  content: attr(title);
```
vinyll: Whaaaatt??? Oh My holly F*** G**! I had no clue we could do that in CSS!
💣

vinyll (approved):

cafebabel/static/css/styles.css
johangiraud: haha ! 🚀

PR #282, opened 2018-03-06
Deal with 404 and 500 errors, fix #131

### LINKED ISSUE #131: #US28 - Gérer les 404 et 500
ETQ lectrice je souhaite consulter une page d'erreur sympathique pour ne pas perdre ma relation avec CaféBabel quand un truc déconne.

- [x] - doit afficher une page pour les erreurs 404 
- [x] - doit afficher une page générique pour les autres erreurs - (400, 403, 500, ...)
- [x] - doit préserver l'URL de requète
- [ ] - pourrait sauvegarder un contexte simplifié dans un fichier consultable : (url, date, erreur)

davidbgk:


davidbgk:
@cafebabel-webmaster I let you the styling part from newly created templates, beware of the comment at the top and avoid using existing templates/macros/whatever.

johangiraud:
Why shouldn't I use existing template actually? It's a bit blocking... How do I add the menu for instance? Or can I use at least the existing stylesheet ? @davidbgk

johangiraud:
While rereading your comment, I think maybe I misunderstood something. You don't want me to use template/macro/whatever everywhere in the 404 and 500 template? Or juste in the top line?

davidbgk:
@cafebabel-webmaster it's a bit painful but necessary, we cannot afford to raise an error because of that page rendering and menus, etc are not error prone. It doesn't have to match the style exactly and you can embed the styles within the `head` section directly if you can, that will avoid static URL issues too.

johangiraud:
ah merde, haven't seen it before commit. It is painful. Ok I'll do that then

cafebabel/__init__.py
```diff
@@ -50,6 +51,32 @@ def register_extensions(app):
...
+            return
+        g.lang = values.pop('lang', app.config['DEFAULT_LANGUAGE'])
+
+
+def register_errors(app):
+    @app.errorhandler(HTTPStatus.NOT_FOUND)
+    def error_not_found(error):
+        return render_template('404.html', error=error), HTTPStatus.NOT_FOUND
+
+    @app.errorhandler(HTTPStatus.INTERNAL_SERVER_ERROR)
```
vinyll, line 74: I always wonder how pertinent this is that the app that is crashing attempts to display an error 500…
I guess we can go that road (much more convenient) and see in the mid-term.

cafebabel/static/css/styles.css
```diff
@@ -2067,3 +2067,28 @@ footer p {
   top: 50%;
   left: 50%;
 }
+
+/* Error pages 404 & 500 */
```
vinyll: Also pinging @davidbgk on this: Shouldn't the page 500 also have it's own css within the page for the sake of not raising some unexpected error when importing an external file?

cafebabel/templates/404.html
```diff
@@ -0,0 +1,31 @@
+{# Beware of not adding any external dependencies here! #}
```
vinyll, line 1: We don't mind on error 404, do we?

cafebabel/templates/500.html
```diff
@@ -0,0 +1,31 @@
+{# Beware of not adding any external dependencies here! #}
```
vinyll, line 1: I would also consider avoiding `url_for` and `include` in this file.
That would mean hardcoding the paths and images url tho.

cafebabel/tests/test_base.py
```diff
@@ -46,7 +46,24 @@ def test_homepage_contains_categories(app, client, published_article):
...
             f'{published_article.author.profile.name}</a>' in response)
+
+
+def test_error_not_found(client):
+    response = client.get('/foobar/')
+    assert response.status_code == HTTPStatus.NOT_FOUND
+    assert '404 error' in response
+
+
+def test_error_internal_server_error(app, client):
```
vinyll, line 61: Oh nice! 👍

cafebabel/templates/404.html
johangiraud: safety is safer, isn't it? @davidbgk

cafebabel/templates/500.html
johangiraud: yep: https://github.com/cafebabel/cafebabel.com/pull/282/commits/97b39cdd8ad8fcc1fa348738be2db026fdf6c882

vinyll:
I feel like it's necessary to extract everything into the 500 or to avoid using `url_for` and `include` (again, in the 500).
I don't feel like we have to for the error 404 nevertheless.
We could do it anyways but it's twice the pain (as the URLs for the images will need to be rewritten when changing environement).
Thoughts?

johangiraud:
In the other hand, we'd need to write the few 404 page css code twice @vinyll.  I'd say we leave it like it is now. I checked a few websites, severals use the usual stylesheet but github don't for example. what do you think? @davidbgk

cafebabel/templates/404.html
```diff
@@ -0,0 +1,91 @@
...
+    text-size-adjust: 100%;
+    -webkit-font-smoothing: antialiased;
+    -moz-osx-font-smoothing: grayscale;
+    font-smoothing: antialiased;
+  }
+  * {
+    box-sizing: border-box;
+  }
+  @font-face {
+    font-family: 'vremena_grotesk';
```
davidbgk, line 19: Do we really need a dedicated font for that page?

cafebabel/templates/404.html
```diff
@@ -0,0 +1,91 @@
...
+    border: 0.1rem solid #00283c;
+  }
+</style>
+
+
+<a href=https://cafebabel.com class=logo>
+  <img src=/static/img/logo-cafebabel.svg alt="logo Cafébabel">
+</a>
+<div id=content>
+  <img src=/static/img/preloader-A.gif alt="coffee grinder gif">
```
davidbgk, line 87: Do we need a preloader to an error page?

davidbgk:
I think it's safer to have as less as we can dependencies for these pages.

cafebabel/templates/404.html
johangiraud: this is not a preloader, this is a gif which is called preloader (because it is also used as a preloader). For now I will put this gif, don't have the time to work further on that

cafebabel/templates/404.html
johangiraud: it's a always the same question with graphic design: do we _really_ need this or that. The answer is no, we dont _really_ need it. But it's better, more consistant and it _really_ does matter. So what do you think?

johangiraud:
Ok @davidbgk let's leave it this way, even if we have to edit 2 template files (`404.html` and `500.html`) when we need to change something

johangiraud (approved):

vinyll (approved):

PR #305, opened 2018-03-07
styling create translation page, fix #273

### LINKED ISSUE #273: US#39 - Étoffer l'édition des traductions
ETQ traducteurs je dois pouvoir éditer un article afin de proposer une traduction

- [ ] doit permettre d'accéder aux fonctionnalités d'éditions équivalentes à la création d'un article

johangiraud:


cafebabel/templates/articles/translations/create.html
```diff
@@ -27,3 +30,8 @@ <h1>Creating a translation in {{ language }} for article “{{ article.title }}
   <input type=submit value="Create">
 </form>
 {% endblock %}
+
+
+
```
davidbgk: That's a lot of empty lines ;)

cafebabel/templates/articles/translations/create.html
```diff
@@ -5,16 +5,19 @@
...
   <input type=hidden name=original value="{{ article.id }}">
-  <div class=text>
+  {% if article.has_image %}
+    <img src="{{ article.image_url }}" alt="Image for {{ article.title }}">
+  {% endif %}
+  <h1 class="edit active">
     <label for=title>Title</label>
-    <input id=title name=title required>
-  </div>
+    <input id=title name=title required value="Translation in {{ language }} for article “{{ article.title }}”">
```
davidbgk: Isn't it a placeholder? I mean do we want the translation to be saved with such a title?

cafebabel/templates/articles/translations/create.html
johangiraud: 😱

cafebabel/templates/articles/translations/create.html
johangiraud: it is a placeholder. And we don't want it to be saved like that. Do you think it's a bad idea to leave it like that?

davidbgk: I think using `placeholder=` is more appropriated then?

johangiraud: I thought we prohibited `placeholders` because of accessibility, didn't we?

davidbgk: It's not the same to use it as a label and to use it as a… placeholder :)

johangiraud: ok ! go for the placeholder then ! https://github.com/cafebabel/cafebabel.com/pull/305/commits/ab2e14a661d7275b244a8e1a6e2d14424d7f04be

cafebabel/templates/articles/translations/create.html
```diff
@@ -5,16 +5,19 @@
   {{ extra_js(['js/articles-tags.js', 'js/textarea-change.js']) }}
 {% endblock %}
 
+{% block body_class %}create-article-page{% endblock %}
```
vinyll: A little confused by the `body_class` names:
`{% block body_class %}profile-page edit{% endblock %}`
`{% block body_class %}tag-page-edit{% endblock %}`
`{% block body_class %}create-article-page{% endblock %}`

It would be less error prone to be consistent here and keep the same name structure.

cafebabel/templates/articles/translations/create.html
```diff
@@ -5,16 +5,19 @@
...
   <input type=hidden name=original value="{{ article.id }}">
-  <div class=text>
+  {% if article.has_image %}
+    <img src="{{ article.image_url }}" alt="Image for {{ article.title }}">
+  {% endif %}
+  <h1 class="edit active">
     <label for=title>Title</label>
-    <input id=title name=title required>
-  </div>
+    <input id=title name=title required value="" placeholder="Translation in {{ language }} for article “{{ article.title }}”">
```
davidbgk: Isn't `value=""` implicit?

cafebabel/templates/articles/translations/create.html
johangiraud: gute Idee ! I replaced `profile-page edit` https://github.com/cafebabel/cafebabel.com/pull/305/commits/6bddf4af7958390f7ad8a9b05f49b3b7e37b42c5 but I don't see how to be more consistent with `tag-page-edit` and  `create-article-page`. They name what the mean, don't they?

cafebabel/templates/articles/translations/create.html
johangiraud: oh right https://github.com/cafebabel/cafebabel.com/pull/305/commits/37046f6269a255a57d1559d57d6152384775ae29

johangiraud:
https://github.com/cafebabel/cafebabel.com/pull/305/commits/89cb47e128f08a675a0652dd52031b1d5c4fdf66 here i tried to use `background-image` instead of `` to avoid image proportion issues. I know inline css is baaad, but I don't know how to do it else. What do you think of that? @vinyll @davidbgk

cafebabel/templates/articles/translations/create.html
johangiraud: Ah sorry, no you are right, in my head we had `create-article-page` AND `article-page-edit` but no 😝 I changed to `article-page-edit` then https://github.com/cafebabel/cafebabel.com/pull/305/commits/362ef772671bed58bab3b4407f31c65ab4df080d

johangiraud:
LGTM except we miss to handle tags. For now I didn't put any field. If I do so, or I get `The fields "{'tag-new', 'tag-2', 'tag-1'}" do not exist on the document "Article.Translation"` if I leave the existing tag (from the original article) or `The field "{'tag-new'}" do not exist on the document "Article.Translation"` if I remove them before

johangiraud:
> LGTM except we miss to handle tags. For now I didn't put any field. If I do so, or I get The fields "{'tag-new', 'tag-2', 'tag-1'}" do not exist on the document "Article.Translation" if I leave the existing tag (from the original article) or The field "{'tag-new'}" do not exist on the document "Article.Translation" if I remove them before

That's the problem I was talking about. Can you help @davidbgk ?

davidbgk:
@cafebabel-webmaster alright I'll take a look at that later this day 👍

johangiraud:
Ok

cafebabel/templates/articles/to-translate.html
```diff
@@ -1,5 +1,6 @@
 {% extends "base.html" %}
 {% import "articles/macros.html" as macros %}
+{% from "articles/macros.html" import card %}
```
davidbgk: Macros are already imported, you can reuse `macros.card()` directly.

johangiraud: I think I had an error without that. I wouldn't have thought about import it again otherwise. Will have a look closer on Monday

davidbgk:
> That's the problem I was talking about.

@cafebabel-webmaster there is a problem before that to me in the way we [select the language](https://github.com/cafebabel/cafebabel.com/blob/master/cafebabel/static/js/articles-tags.js#L9), we need to rethink that part a bit on how we select the current language from the template.

johangiraud:
> @cafebabel-webmaster there is a problem before that to me in the way we select the language, we need to rethink that part a bit on how we select the current language from the template.

Ok let's discuss it today then

cafebabel/templates/articles/to-translate.html
johangiraud: oui voilà, I get `jinja2.exceptions.UndefinedError: 'card' is undefined` otherwise. I guess `import` was not the right way.
https://github.com/cafebabel/cafebabel.com/pull/305/commits/7287ca0ebba14fb2da5410c6ba19cdc3f6937090

davidbgk: you need to prefix `card` like that: `macros.card`, did you do that?

johangiraud: https://github.com/cafebabel/cafebabel.com/commit/7287ca0ebba14fb2da5410c6ba19cdc3f6937090 isn't it fine this way? with `{% from "articles/macros.html" import card %}` ?

cafebabel/templates/articles/macros.html
```diff
@@ -18,7 +18,7 @@ <h4>Words by:</h4>
       {% for author in article.authors %}
           <a href="{{ url_for('users.detail', id=author.pk) }}" class=author-link>
             {% if author.profile.has_image %}
```
davidbgk: I would create a `author_image(author)` macro dedicated to that repeated case.

cafebabel/templates/articles/translations/create.html
```diff
@@ -5,16 +5,19 @@
...
   <input type=hidden name=original value="{{ article.id }}">
-  <div class=text>
+  {% if article.has_image %}
+    <img src="{{ article.image_url }}" alt="Image for {{ article.title }}">
+  {% endif %}
+  <h1 class="edit active">
     <label for=title>Title</label>
-    <input id=title name=title required>
-  </div>
+    <input id=title name=title required placeholder="Translation in {{ language }} for article “{{ article.title }}”">
```
davidbgk: Using it with real data, the whole line is hardly visible given the size of the text. Maybe we could put that indication somewhere else where it's kept during edition (above or below the field)?

cafebabel/templates/articles/to-translate.html
davidbgk: @cafebabel-webmaster if you only use that macro that's the way to go 👍

cafebabel/templates/articles/macros.html
johangiraud: ok: https://github.com/cafebabel/cafebabel.com/pull/305/commits/1e97685adf27432cb9f69d50180d84e62d4f925f

cafebabel/templates/articles/translations/create.html
johangiraud: #244 ? not a priority I assume

cafebabel/templates/articles/to-translate.html
johangiraud: 😃

davidbgk:
I have an issue related to the `to-translate` page, that's not the first time I cannot understand if the Translate button is associated with the article mentioned above or below, maybe add extra space for better grouping of related elements? Or maybe it's just me.

johangiraud:
> I have an issue related to the to-translate page, that's not the first time I cannot understand if the Translate button is associated with the article mentioned above or below, maybe add extra space for better grouping of related elements? Or maybe it's just me.

better? https://github.com/cafebabel/cafebabel.com/pull/305/commits/d47d431f81fbe6ff99cdcb363ec298fa2df5767a

davidbgk:
> better?

Still not that clear to me, which article that button belongs to?

capture d ecran 2018-03-12 a 14 07 35

cafebabel/static/css/styles.css
```diff
@@ -2175,10 +2189,14 @@ a.section-link.creative {
   margin-bottom: 5rem;
 }
 
-.home article .article-detail h3 {
+article .article-detail h3 {
   font-size: 2.2rem;
 }
 
+article .article-detail h3 a:hover:after {
```
vinyll, line 2196: Can you double check that `:hover:after` works? I would have written `:after:hover`

cafebabel/templates/articles/macros.html
```diff
@@ -8,6 +8,14 @@
   {% endif %}
 {% endmacro %}
 
+{% macro profile_image(author) %}
```
vinyll, line 9: 👌

cafebabel/templates/articles/macros.html
```diff
@@ -8,6 +8,14 @@
...
+  {% if author.profile.has_image %}
+    <div style="background-image:url({{ author.profile.image_url }})" class=image></div>
```
vinyll: If I remember well, you do this to be able to resize into a centered squared with preserved ratio.
As this is hackish, could you comment in the code _why_ you need to put the image in background instead of ``?

cafebabel/templates/users/detail.html
```diff
@@ -1,18 +1,14 @@
 {% extends 'base.html' %}
 {% from "users/macros.html" import social_field, default_profile_picture with context %}
```
vinyll: `default_profile_picture` isn't used anymore

cafebabel/templates/users/edit.html
```diff
@@ -1,12 +1,12 @@
 {% extends 'base.html' %}
 {% from "users/macros.html" import social_field, default_profile_picture with context %}
```
vinyll: same here

cafebabel/templates/articles/macros.html
```diff
@@ -8,6 +8,14 @@
...
+  {% else %}
+    {{ default_profile_picture() }}
```
vinyll: the macro `default_profile_picture` is now only used here.
Therefore it should be deleted and it's code moved here.

cafebabel/templates/articles/translations/edit.html
```diff
@@ -1,18 +1,38 @@
 {% extends "base.html" %}
+{% import "articles/macros.html" as macros %}
 {% from "_extrajs-macros.html" import extra_js %}
 
 {% block extra_js %}
-  {{ extra_js(['js/articles-tags.js', 'js/textarea-change.js']) }}
+  {{ extra_js(['js/articles-tags.js', 'js/textarea-change.js', 'js/noise.js']) }}
```
vinyll: I can't find where noise is used in this code. Any hint @cafebabel-webmaster?

vinyll (changes requested):

vinyll:
@cafebabel-webmaster https://github.com/cafebabel/cafebabel.com/pull/305#issuecomment-372165949 was fixed in #338

cafebabel/templates/articles/translations/edit.html
davidbgk: Agreed, no cover image modification is allowed from that page.

johangiraud:
> Still not that clear to me, which article that button belongs to?

@davidbgk you are not looking into the right branch I guess.

capture d ecran 2018-03-13 a 10 15 07

cafebabel/static/css/styles.css
johangiraud: It does work? I used it lines 643, 673, 851, 875, 923, 991, 1478, 1861, 2039, 2156 etc. ;-)

cafebabel/templates/articles/macros.html
johangiraud: yep

cafebabel/templates/users/detail.html
johangiraud: ok

cafebabel/templates/articles/macros.html
johangiraud: @davidbgk did it here https://github.com/cafebabel/cafebabel.com/pull/305/commits/3a47e23c64d4a0bb85ba22d3212b235118d2e87c

cafebabel/templates/articles/translations/edit.html
johangiraud: true, user can't change cover image at this point. Then I delete it

johangiraud:
in to-translate page, I have:
```
{% for article in articles %}
    {% if not article.is_translated_in(to_language_code) %}
      <li>
        {{ card(article) }}
          {% if article.original_article.language != to_language_code %}
            <a class=button href="{{ url_for('translations.create') }}?lang={{ to_language_code }}&original={{ article.original_article.id }}">Translate in {{ to_language_label }}</a>
          {% endif %}
            {% if article.original_article.language != to_language_code %}
              <a class=button href="{{ lang_url_for('translations.create') }}?lang={{ to_language_code }}&original={{ article.original_article.id }}">Translate in {{ to_language_label }}</a>
            {% endif %}
          {% else %}
            <a class=button href="{{ lang_url_for('translations.create') }}?lang={{ to_language_code }}&original={{ article.id }}">Translate in {{ to_language_label }}</a>
          {% endif %}
      </li>
    {% endif %}
  {% endfor %}
```
First I got an error: `cafebabel.articles.models.Article object` and second I think we want only translations from the original article. Is this code relevant, then?

johangiraud:
> @cafebabel-webmaster #305 (comment) was fixed in #338

@vinyll it still doesn't work if I enable tags in a translation: `The fields "{'tag-1', 'tag-new'}" do not exist on the document "Article.Translation"`

cafebabel/templates/articles/translations/create.html
```diff
@@ -8,24 +9,25 @@
...
+    <div class=textarea>
+      <label for=summary>Teaser</label>
+      <textarea id=summary name=summary required></textarea>
+    </div>
+    <div class=textarea>
+      <label for=body>Body</label>
+      <textarea id=body name=body required></textarea>
+      {% include "articles/markdown-helper.html" %}
+    </div>
+    {{ macros.tags_fields(translation) }}
```
johangiraud, line 30: Here i screwed up, I juste add this line actually: `{{ macros.tags_fields(translation) }}`. The rest stayed the same @davidbgk

johangiraud:
Can you help me fixing the broken test @davidbgk? I can't do it 😞

davidbgk:
@cafebabel-webmaster I'll take care of it ;)

johangiraud:
merci !

cafebabel/templates/articles/translations/edit.html
```diff
@@ -42,6 +42,7 @@ <h1 class="edit active">
     <textarea id=body name=body required>{{ translation.body }}</textarea>
     {% include "articles/markdown-helper.html" %}
   </div>
+  {{ macros.tags_fields(translation) }}
```
vinyll: I suppose that if we add it here we also want it for translation/create?

johangiraud:
LGTM

cafebabel/tests/test_translations.py
```diff
@@ -82,6 +80,50 @@ def test_translation_creation_should_redirect(app, client, user, article):
...
+def test_translation_creation_with_existing_tag(app, client, user, article):
+    login(client, user.email, 'password')
+    language = app.config['LANGUAGES'][1][0]
+    tag = Tag.objects.create(name='Sensational', language=language)
+    data = {
+        'title': 'title',
+        'summary': 'summary',
+        'body': 'body',
+        'original': article.id,
+        'language': language,
```
vinyll, line 93: `'language': 'fr',` to make sure and obvious that browsing `/fr/…` will match as expected.

cafebabel/tests/test_translations.py
```diff
@@ -82,6 +80,50 @@ def test_translation_creation_should_redirect(app, client, user, article):
...
+
+def test_translation_creation_with_unknown_tag(app, client, user, article):
+    login(client, user.email, 'password')
+    language = app.config['LANGUAGES'][1][0]
+    data = {
+        'title': 'title',
+        'summary': 'summary',
+        'body': 'body',
+        'original': article.id,
+        'language': language,
```
vinyll, line 115: Same here I guess

vinyll (approved):
r+wc

cafebabel/tests/test_translations.py
davidbgk: The `language` is in use also to reference the tag creation, I would keep it as is for consistency, thoughts?

Or maybe I should use `{language}` within the URL to uniformize?

vinyll: I guess your suggestion of `{language}` in the url would make sense to me.
That's not a blocker though, just a consistency matter.

vinyll:
LGTM

PR #323, opened 2018-03-07
Translate from current language

### LINKED ISSUE #308: US #38 - /to-translate/ should only offer destination language
When offering to participate to a translation (en > fr by example), I am on cafebabel.com/en/.
Therefore only the articles in English are displayed.

I should only be offer to translate this article into a selection of languages that is still available, and from the current browsing language.

vinyll:
fix #308


vinyll:
I should first reread this before getting this read by someone else.
But taking the ✈️

vinyll:
It was decided in daily — after asking the translator team — that translating from the current language to a foreign language was too restrictive.
Therefore this PR is being closed

PR #333, opened 2018-03-12
WIP: Page transitions

johangiraud:
Ref to #173 
I don't know how much "dirty" it is, but I have it work like that so far.
WIP 2 errors:
- Possible different images called when clicking on a link AND before loading the page
- Sometimes, preloader images are not loaded fast enough (I guess), and only the alt text is displayed


davidbgk:
I really don't get why we need preloader(s), where is the associated issue? Which is the need? Where is the US for this sprint?

johangiraud:
it's not in the sprint. I didn't ask for any review though.
We need preloader(s) because it makes the navigation more fluid and smooth.

PR #338, opened 2018-03-12
tags selector use current language

### LINKED ISSUE #337: Tag selector relays on #language dom selector
As we don't select the language anymore, [this code](https://github.com/cafebabel/cafebabel.com/blob/master/cafebabel/static/js/articles-tags.js#L9) is outdated

vinyll:
fix #337


davidbgk (approved):
😘

PR #340, opened 2018-03-12
Ensure all url_for are localized, fix #299

### LINKED ISSUE #299: US#38 Localize each and every link
Refs https://github.com/cafebabel/cafebabel.com/issues/221

Check every `url_for` call.

davidbgk:
In the meantime, switch back to ids from pks…


vinyll:
Why not implementing a `lang_url_for` instead that would take the current lang instead of demultiplying the `lang=current_language`?

vinyll (approved):

PR #346, opened 2018-03-13
Change language on categories redirects, fix #297

### LINKED ISSUE #297: US #38 - changing language on a category displays the equivalent category
when browsing on `/en/tag/impact/` and changing to fr, I should be directed to `/fr/tag/impact/`.
Same for the other 3 categories.

#290

davidbgk:
Note that no check is performed to check if the target category exists prior to make the link available, it is intentional as it is costly and all categories should be present for all languages in a real case scenario.


vinyll (approved):
Super clean 💅

PR #368, opened 2018-03-15
prod deployment

vinyll:
Note that the domain _cafebabel.com_ is not set yet.
You will need to `sudo  sh -c 'echo "185.34.32.17 cafebabel.com" >> /etc/hosts'` to access the website.

You may also need to empty you URL cache for _cafebabel.com_ if you had browsed that domain before (or you will be redirected to _www_ as with the old system).


Makefile
```diff
@@ -1,11 +1,12 @@
 # Config
-server=preprod@185.34.32.17
+server=${env}@185.34.32.17
 src_dir=~/cafebabel.com
 venv_dir=~/cafebabel.com/venv
 branch=master
-LOGS=access
 old_prod=prod@91.194.60.65
 old_prod_media_dir=~/cafebabel/data/medias
+prod_server=127.0.0.1:5001
```
davidbgk, line 8: I guess this was for local tests?

Makefile
```diff
@@ -15,44 +16,40 @@ goto_src=cd ${src_dir} && source ${venv_dir}/bin/activate
...
-	make rsync
-	@echo "> Launching gunicorn daemon."
-	${remote} "${goto_src} && pkill gunicorn; \
-		gunicorn --daemon -b 127.0.0.1:5000 preprod:app \
-		--error-logfile ~/log/preprod.log --access-logfile ~/log/preprod.log"
-	@echo "> App is deployed. Run \`make logs\` to follow activity."
-
-rsync:
-	@echo "> Synchronizing archives media from old production server."
+	make sync_archives_media
```
davidbgk: I tried to keep consistent with dashes in command name, bad idea?

Makefile
```diff
@@ -15,44 +16,40 @@ goto_src=cd ${src_dir} && source ${venv_dir}/bin/activate
...
 	-${remote} "python3.6 -m venv ${venv_dir}"
 	-${remote} "${goto_src} && make make-dirs"
-	make reset-db
 	make deploy
 
 make-dirs:
 	mkdir -p ./logs
+	mkdir -p ./cafebabel/uploads/{archives,articles,tags,users}
 
-reset-db:
```
davidbgk: If we remove this, we have to document Flask commands to launch for setting up the database?

Makefile
vinyll: No, this is the server running in production with gunicorn

Makefile
vinyll: Good to me. I was waiting for your other PR to be merged to minimize conflicts in name.

Makefile
vinyll: Right, let's keep it

Makefile
davidbgk: Oh I just noticed the `${${env}_server}` 😈

Makefile
davidbgk: Can you do it?

davidbgk (approved):
r+wc

Makefile
vinyll: Oups, I had in mind I had pushed that 😅

PR #372, opened 2018-03-15
Wording

johangiraud:


cafebabel/templates/articles/proposals/create.html
```diff
@@ -42,11 +46,11 @@ <h3>Wanna be part of Cafébabel project? That’s the right place</h3>
           </select>
         </label>
       </div>
-    <label for=field_additional>Additional informations</label><textarea name=additional id=field_additional></textarea>
+    <label for=field_additional>Additional information</label><textarea name=additional id=field_additional></textarea>
 
-    <button type=submit>Suggest this article</button>
+    <button type=submit>You're all done! Submit your pitch</button>
```
davidbgk: I'd use a `’` instead of the `'` to be consistent.

davidbgk (approved):
r+wc

cafebabel/templates/articles/proposals/create.html
johangiraud: ok !

cafebabel/tests/test_articles.py
```diff
@@ -370,10 +370,10 @@ def test_article_to_translate_should_have_translation_links(
     article.modify(language=language)
     response = client.get(f'/en/article/to-translate/?from=fr&to=en')
     assert (f'href="/en/article/translation/new/'
-            f'?lang=en&original={article.id}">Translate in English</a>'
+            f'?lang=en&original={article.id}">Translate into English</a>'
```
johangiraud: Sorry @davidbgk my page hadn't been refreshed, so I fixed the tests a second time :-(

cafebabel/templates/articles/translations/create.html
```diff
@@ -16,18 +16,18 @@
     {% endif %}
     <h1 class="edit active">
       <label for=title>Title</label>
-      <input id=title name=title required placeholder="Translation in {{ language }} for article “{{ article.title }}”">
+      <input id=title name=title required placeholder="{{ article.title }}">
```
davidbgk, line 19: I'm getting back at this 🙈 

To me a placeholder is to give an example of what could be typed, here I think the title of the original article is valuable to stay during edition to be able to translate it while seeing it. Proposition: putting it within the `label` like that `<label for=title>Title (original: {{ article.title }})</label>` maybe even with a link to the article in order to be able to access it quickly given there is no text to translate on that page.

johangiraud: hum it seems to me more confusing to have the original title as a `label`. I will leave it like that for now, and wait for feedback from users. But thanks for your proposition 😉

johangiraud:
@davidbgk il reste juste un cas de figure où on a la phrase _Your article was successfully updated._ c'est quand on a un article publié et qu'on le passe en draft. Désolé d'être picky mais tu peux changer ça please ?

davidbgk:
@cafebabel-webmaster I don't get it, there is no occurrence of that string in the code

davidbgk:
@cafebabel-webmaster and we have no option to give that particular message on status change (as previously discussed)

johangiraud:
hum yes, that's why I was wondering... strange

EDIT : OK it's working, my fault, I hadn't pulled on the right branch

PR #373, opened 2018-03-15
Do not set a publication date for drafts, fix #348

### LINKED ISSUE #348: Consulter les drafts par la langue sélectionnée
**Page** : view Drafts
**Contexte** : je consulte les drafts avec une langue sélectionnée.
**Préconisation** : je souhaiterais que ne soit affiché que les drafts de la langue sélectionnée

davidbgk:
As a side effect, new draft would not appear on the list page because they were preceded by ones with a publication_date

Warning: requires a new import of data from old prod.

### DIFF
--- a/cafebabel/django_migrations.py
+++ b/cafebabel/django_migrations.py
@@ -205,6 +205,8 @@ def create_article(old_article):
             relateds=fields['relateds']
         ),
     }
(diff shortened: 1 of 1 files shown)


vinyll (approved):
Bien vu ! 😅

PR #378, opened 2018-03-15
center article cover picture if they are less than 1200px wide

johangiraud:


davidbgk (approved):
I guess it will center all images in article and you are aware of that(?)

johangiraud:
No it will only center the img which is right after an `<article>` tag inside a container `.article-page` 😉

johangiraud:
There wasn't any associated issue @pointbar. Was a mystery why it hasn't been deployed yet. I thought we did it months ago.. sorry for that.

PR #383, opened 2018-03-16
adapt tag-list styles to avoid interferences with regular article pin…

johangiraud:
…k styles

### DIFF
--- a/cafebabel/static/css/styles.css
+++ b/cafebabel/static/css/styles.css
@@ -2004,6 +2004,7 @@ article .article-detail p span {
 .tags-list a {
   position: relative;
   font-size: 1rem;
+  font-weight: 200;
   padding: 0.3rem 0.5rem;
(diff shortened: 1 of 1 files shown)


cafebabel/static/css/styles.css
```diff
@@ -2029,17 +2030,25 @@ article .article-detail p span {
   content: '#';
 }
 
-.tags-list a.impact,
-.tags-list a.experience,
-.tags-list a.raw,
-.tags-list a.creative {
+.article-page article p.tags-list a.impact,
```
vinyll, line 2033: Is `.article-page .tags-list a.impact` not specific enough?

johangiraud: no :-( cause it gets overridden by `.article-page article p.tags-list a, .tags-list a` line 2004

vinyll (approved):

PR #391, opened 2018-03-29
Article image arrangements

johangiraud:


cafebabel/static/css/styles.css
```diff
@@ -1554,6 +1554,7 @@ select#author {
 
 .article-page article > img {
   margin: 0 auto;
+  width: 100%; /* force archive images (1180px wide) to be 1200px */
```
davidbgk, line 1557: Isn't it more pertinent to keep 1180px to avoid blurriness?

johangiraud: First I hate this random 1180 number (not a round number: quite fussy I know), and second I set the content max-width to 1200px everywhere cause it's very common and it makes more sense to me as well. What do you think? I know it's very bad to stretch images, but for 20px, I don't care. Really.

johangiraud:
Ok! Both team and me agreed with stretching-1180px-images-to-1200px solution 🎉  can you approve the PR please so that I can merge? @davidbgk or @vinyll

davidbgk (approved):

PR #392, opened 2018-03-29
change 'auto' to '0' to prevent from jumping when editing, fix #280

### LINKED ISSUE #280: Return button pressed inside text body makes cursor jump to end of page
**Page**: /article/edit
**Contexte**: 
Sur un article composé de plus de 60 lignes, quand j'ajoute un caractère après la ligne 56.

**Problème**:
Ca fait sauter le curseur à la fin de la page (le footer).

johangiraud:
fix #280 
problem left : textarea aren't resized when cuting or deleting content 😢

### DIFF
--- a/cafebabel/static/js/textarea-change.js
+++ b/cafebabel/static/js/textarea-change.js
@@ -1,7 +1,7 @@
 /* Auto-expand textarea when typing http://jsfiddle.net/hmelenok/WM6Gq/ */
 Array.from(document.querySelectorAll('textarea')).forEach(textArea => {
   function resize() {
-    textArea.style.height = 0 /* textarea is NOT resized when deleting or cut content, to hack it : 'auto' (but it jumps when editing and saving :-( */
+    textArea.style.height = 0 /* textarea is NOT resized when deleting or cut content, to hack it : 'auto' (but it jumps when editing and saving) :-( */
     textArea.style.height = `${textArea.scrollHeight}px`
   }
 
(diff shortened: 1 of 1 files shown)


cafebabel/static/js/textarea-change.js
```diff
@@ -1,8 +1,7 @@
 /* Auto-expand textarea when typing http://jsfiddle.net/hmelenok/WM6Gq/ */
 Array.from(document.querySelectorAll('textarea')).forEach(textArea => {
   function resize() {
-    textArea.style.height =
-      'auto' /* hack to allow textarea to be resized when deleting or cut content */
+    textArea.style.height = 0 /* textarea is NOT resized when deleting or cut content, to hack it : 'auto' (but it jumps when editing and saving) :-( */
```
davidbgk: The comment is misleading, can you explain why there is `0`?

johangiraud: I can't explain why it needed to be `auto` (since it's set instantly to `${textArea.scrollHeight}px`), nor I can explain why `0` fixes the jumping issue but prevent from resizing the textarea when deleting text. We can talk about it orally later, can't we?

johangiraud: Any thoughts @davidbgk or @vinyll ?

davidbgk: Proposal: `/* textarea is NOT resized when deleting or cutting content, to hack it, we tried 'auto' but it jumps when editing and saving so we had to fallback on '0' :-( */`

davidbgk (approved):
See comments.

PR #397, opened 2018-03-30
add specific tag class to body

### LINKED ISSUE #395: Page Network
- Discriminer la page tag network pour pouvoir la styler autrement

johangiraud:
fix #395

### DIFF
--- a/cafebabel/templates/articles/tags/detail.html
+++ b/cafebabel/templates/articles/tags/detail.html
@@ -30,7 +30,7 @@
   {{ meta('name=twitter:url', request.url) }}
 {% endblock %}
 
-{% block body_class %}tag-page {{ tag.slug }}{% endblock %}
+{% block body_class %}tag-page tag-{{ tag.slug }}{% endblock %}
 {% block body %}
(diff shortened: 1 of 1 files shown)


cafebabel/templates/articles/tags/detail.html
```diff
@@ -30,7 +30,7 @@
   {{ meta('name=twitter:url', request.url) }}
 {% endblock %}
 
-{% block body_class %}tag-page{% endblock %}
+{% block body_class %}tag-page {{ tag.slug }}{% endblock %}
```
davidbgk: I think you should prefix the tag to avoid collisions with existing body classes (and associated styles).

davidbgk (changes requested):

cafebabel/templates/articles/tags/detail.html
johangiraud: ok https://github.com/cafebabel/cafebabel.com/pull/397/commits/ad4d0d032fc3c75a005d57f428562f1effb8a601

davidbgk (approved):

PR #399, opened 2018-04-02
Make emails visible to editors on profiles, fix #207

### LINKED ISSUE #207: Issue #US38 - Adresse mail non visible dans le profil
**URL**: profil/detail
**Contexte**: je suis éditrice et je veux trouver l'email d'un contributeur en allant sur son profil
**Problème**: lorsque je vois le _full name_, je ne peux plus voir l'adresse mail du contributeur
**Solution**: ajouter un champs pour l'adresse mail visible uniquement par les éditrices et par le contributeur lui-même

davidbgk:
And also to the current logged in user.

The email is clickable for now. No more default username fallback to email so we don’t leak it. Default set to `Anonymous` because they are legion!

### DIFF
--- a/cafebabel/templates/users/detail.html
+++ b/cafebabel/templates/users/detail.html
@@ -4,7 +4,7 @@
 
 {% block body_class %}profile-page{% endblock %}
 {% block body %}
-<h1>{{ user }}</h1>
+<h1>{{ user.profile.name or '' }}</h1>
 
 <section>
   <h2>About</h2>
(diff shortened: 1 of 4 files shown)


cafebabel/templates/users/detail.html
```diff
@@ -22,7 +22,12 @@ <h2>About</h2>
   {% endif %}
 
   <ul class=info-profile>
-    <li><span>{{ user.profile.name or '' }}</span></li>
+    <li>
+      <span>{{ user.profile.name or '' }}</span>
+      {% if user.is_me() or current_user.has_role('editor') %}
+        / <span><a href="mailto:{{ user.email }}">{{ user.email }}</a></span>
```
vinyll: Why the prepending `/`?

davidbgk: The separator is left at @cafebabel-webmaster discretion ;)

johangiraud:
For now it means users can't change their email address? Me neither btw. I thought it would have been more convinient to have a proper field for that

johangiraud:
> No more default username fallback to email so we don’t leak it. Default set to Anonymous because they are legion!

I think we should actually ask for full name when registering so that we don't have `Anonymous`. I see it coming: users will never set any full name and it will be such a pain for editors to assign contributors to articles cause there will be planty of `Anonymous` profiles :/

johangiraud (changes requested):
Shouldn't we add a field for email address for users to change it if they need?

vinyll:
2 issues have been open: #402 and #403

vinyll (approved):

johangiraud (approved):

PR #401, opened 2018-04-02
Do not allow translations from translations, fix #274

### LINKED ISSUE #274: Traduction à partir d'une traduction impossible
ETQ traducteur je ne veux pas pouvoir faire de traduction à partir d'une traduction existante afin de préserver la consistance de l'article

- [x] ne doit afficher que les traductions possibles à partir de la langue racine

davidbgk:


davidbgk:
@cafebabel-webmaster I set you as a reviewer mainly to check with editors if it's now pertinent for editors, not sure if that should be @pointbar job but it's probably easier being close to them 😉

johangiraud (changes requested):
If I visit a non-original-language article, I can still create a translation by hitting a select-language button: 
capture d ecran 2018-04-03 a 11 49 10

vinyll:
@cafebabel-webmaster: the translations are based on the original article. See the translation id `…a516` and the original article that the translation will be based on `…a515`

screen shot 2018-04-03 at 12 43 55 pm

johangiraud:
> @cafebabel-webmaster: the translations are based on the original article. See the translation id …a516 and the original article that the translation will be based on …a515

ah ok, looking good to me then!

davidbgk:
> ah ok, looking good to me then!

Can I get a new review then? :)

vinyll (approved):

johangiraud (approved):

PR #425, opened 2018-04-10
removed jinja comment on error 500 display

vinyll:
When Nginx is rendering the error 500, the jinja comment appeared at the top of the page.


davidbgk (approved):
👍

PR #430, opened 2018-04-12
editing a published translation cannot edit contributors

### LINKED ISSUE #424: Remove the ability to edit authors/translators on translations
Not consistent and is broken as of now.

vinyll:
fix #424


cafebabel/tests/test_translations.py
```diff
@@ -326,7 +326,7 @@ def test_translation_published_should_have_translator(
              f'article title') in response)
     assert (f'<a href="/fr/profile/{translator.id}/" class=translator-link>'
             in response)
-    assert f'{translator}' in response
+    assert translator in response
```
davidbgk: Tests are failing because of this, maybe `str(translator) in response` would be enough.

vinyll: Funny that's what I just pushed ;)

davidbgk: You sure?!

vinyll: Yes but the pushed had been rejected after master update :/

davidbgk (approved):
Maybe indent the `if` in template now the review is done?

PR #484, opened 2018-05-09
Header arrangements

johangiraud:


johangiraud:
Here I have a js error @vinyll in common.js. It's about this line:
```
const autofillContent=`"${String.fromCharCode(0xfeff)}"`function checkAutofill(input){if(input&&!input.value){const style=window.getComputedStyle(input)
```
I don't know how it is compiled, but I see this comment: `// Keep semicolon for jsmin.` So I guess prettier.js removes a semicolon, maybe that could be the problem. How to force prettier.js to leave the semicolon?

johangiraud:
Ok after trying many times to leave the semicolon while saving, I only just didn't save, then commit and it worked! 🎉 Not very handy when I will have to edit this file (`script.js`) but it will do for now 😃

PR #489, opened 2018-05-14
Profile-edit page layout

johangiraud:


johangiraud:
`translation` is undefined in _users/edit.html_. How can I define it? @vinyll @davidbgk

johangiraud:
Here I need help to close :

>`translation` is undefined in users/edit.html. How can I define it? @vinyll @davidbgk

johangiraud:
I guess it just needs to be declared somewhere in the `helper.js` or so? @vinyll

vinyll:
@cafebabel-webmaster: I'm not certain I understand your issue well.
`{% macro contribution_list() %}` should probably `{% macro contribution_list(articles, type) %}`
where `type` can be "translations" or "articles" and `articles` can be a list of translations or of articles.

johangiraud:
No, I need it to be both at the same time: article list and translation list. Exactly the way it works in `user/detail` actually @vinyll

vinyll:
The translations and the articles are not mixed in the HTML.
In users/details.html, you can do:
```html
…
{% macro contribution_list(articles, 'articles') %}
…
{% macro contribution_list(translations, 'translations') %}
…
```

johangiraud:
Ok I try

johangiraud:
No, because in `{{ contribution_list(articles, 'articles' }}` I would need `translations` variable too and the other way around.

vinyll (approved):

vinyll:
You're getting into the next step in Python! 🐍

johangiraud:
>You're getting into the next step in Python! 🐍

Yeah!

PR #496, opened 2018-05-16
Google tag manager

johangiraud:


cafebabel/tests/test_articles.py
```diff
@@ -49,7 +49,7 @@ def test_published_article_should_display_content(client, published_article,
     assert published_article.authors[0].profile.name in response
     assert ('href="https://twitter.com/share?url=http%3A%2F%2Flocalhost%2F'
             f'en%2Farticle%2F{published_article.slug}-{published_article.id}'
-            f'%2F&text={published_article.title}&via=cafebabel_eng"'
+            f'%2F&text={published_article.title}&via={twitter_nickname}"'
```
vinyll: Tu veux garder `via=cafebabel_eng` ici

cafebabel/tests/test_articles.py
```diff
@@ -31,7 +31,7 @@ def test_article_with_non_ascii_title(published_article):
 
 
 def test_published_article_should_display_content(client, published_article,
-                                                  user):
+                                                  user, twitter_nickname):
```
vinyll: `twitter_nickname` is not something that exists.
That breaks the tests

vinyll (changes requested):

cafebabel/tests/test_articles.py
johangiraud: yep, I created a new branch from master, I don't understand why I get this here. My bad probably ;-)

cafebabel/tests/test_articles.py
johangiraud: same

vinyll (approved):

